PR #57, opened 2013-01-31
v3 and v5 UUID support

OrangeDog:


selfcontained:
It's a bummer this died here, a stand-alone uuid-v3 module would be super handy, but I'd hate to just rip this code off and call it my own.

tracker1:
@selfcontained  @broofa  Could use/reference browerify's shim for node's crypto (incomplete, but iirc includes sha1 and md5 bits)

coolaj86:
Any further thoughts on this?

I think it would be fine to put this in a subfolder such that you `require('node-uuid/v3')` or something like that.

jamesdixon:
Bump. Any updates?

PR #88, opened 2014-09-08
Only try to use crypto.randomBytes if env is node (does not work in browser amd)

gaye:


fedot:
This will not work as desired due to issue with `require` detection, see @broofa comment on the original commit that introduce the issue https://github.com/broofa/node-uuid/commit/319dc6e2e75f8ff57e9e708e2187bdabba6ca86a

uuid.js
```diff
@@ -14,7 +14,7 @@
   // Node.js crypto-based RNG - http://nodejs.org/docs/v0.6.2/api/crypto.html
   //
   // Moderately fast, high quality
-  if (typeof(_global.require) == 'function') {
+  if (typeof(_global.require) == 'function' && typeof(module) != 'undefined' && module.exports) {
```
fedot: `_global.require` is always `undefined` in node.js

gaye: No it isn't?

```
> this.require
{ [Function: require]
  resolve: [Function],
  main: undefined,
  extensions: 
   { '.js': [Function],
     '.json': [Function],
     '.node': [Function: dlopen] },
  registerExtension: [Function],
  cache: {} }
```

fedot: Test it outside of node REPL and you'll see a different picture

   (function() {
     var _global = this;
     console.log(_global.require);
   }).call(this);

gaye: Why is that?

coolaj86:
"fixed" in #126

PR #94, opened 2014-10-31
remove unused granularity.

tracker1:
JS Clock is only granular up to milliseconds, so comparing with the counter has no real value... the milliseconds will be equal or not.  should still catch a regression in this case.

### DIFF
--- a/uuid.js
+++ b/uuid.js
@@ -131,7 +131,7 @@
     var nsecs = options.nsecs != null ? options.nsecs : _lastNSecs + 1;
 
     // Time since last uuid creation (in msecs)
-    var dt = (msecs - _lastMSecs) + (nsecs - _lastNSecs)/10000;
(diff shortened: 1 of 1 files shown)


coolaj86:
I think what we ought to do is use `process.hrtime()`

broofa:
See my comments about `hrtime` in #109.  (tl;dr - not worth the hassle)

broofa:
Closing this, as I think there's still a case where nsecs - _lastNSecs may be significant (e.g. when options.nsecs is specified at some point after _lastNSecs being non-zero.

coolaj86:
What's the issue here with the granularity then? Since the clock step is happening, we do want the extra granularity, correct?

Are we safe to close this issue?

coolaj86:
oh, ok

PR #102, opened 2015-02-11
Force to _hexToByte to be fast V8 object

Ginden:
This change results in:
```
C:\Users\michal.wadas\Documents\GitHub\node-uuid\uuid.js:159
      throw new Error('uuid.v1(): Can\'t create more than 10M uuids/sec');
```
on my computer.

### DIFF
--- a/uuid.js
+++ b/uuid.js
@@ -59,6 +59,16 @@
     _hexToByte[_byteToHex[i]] = i;
   }
 
+  // Force _hexToByte to be changed to fast-property object in V8
+  (function toFastProperties(obj) {
+    /*jshint -W027*/
+    function f() {}
+    f.prototype = obj;
+    return f;
+    eval(obj);
+  }(_hexToByte));
+
+
   // **`parse()` - Parse a UUID into it's component bytes**
   function parse(s, buf, offset) {
     var i = (buf && offset) || 0, ii = 0;


uuid.js
```diff
@@ -59,6 +59,16 @@
     _hexToByte[_byteToHex[i]] = i;
   }
 
+  // Force _hexToByte to be changed to fast-property object in V8
+  (function toFastProperties(obj) {
+    /*jshint -W027*/
+    function f() {}
+    f.prototype = obj;
+    return f;
+    eval(obj);
```
broofa: Does this serve a purpose?  Isn't this ...

  1. unreachable
  2. a noop (eval just returns `obj`, right?)

Ginden: Yes, it's unreachable and noop, but it prevents V8 from removing dead code. 
Kind of voodoo.

broofa:
So, here's the thing... from a performance perspective this is awesome and I love it.  But from a maintainability perspective this scares me.  It's (wait for it) "voodoo" code that is nearly impossible to understand.  Unless it serves a clear purpose (as in, solves an real issue some has with performance in a production environment) I'm inclined to reject this in the interests of code clarity.

Thoughts?

coolaj86:
Doesn't sound like there's actually a current performance issue here and if there were, there are [other ways](https://github.com/broofa/node-uuid/issues/125) to speed things up - like moving regex to the top and pooling crypto bytes.

Closing.

PR #107, opened 2015-03-17
Provide support of MSIE 11 msCrypto

danorton2:
MSIE 11 removed support for crypto, renaming their version as msCrypto.

This patch uses window.msCrypto if window.crypto is unavailable (or falls back to the uuid module's RNG).

### DIFF
--- a/uuid.js
+++ b/uuid.js
@@ -11,6 +11,9 @@
   // returns 128-bits of randomness, since that's what's usually required
   var _rng;
 
+  // Allow for MSIE11 msCrypto
+  var _crypto = _global.crypto || _global.msCrypto;
+
   // Node.js crypto-based RNG - http://nodejs.org/docs/v0.6.2/api/crypto.html
   //
   // Moderately fast, high quality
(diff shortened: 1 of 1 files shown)


broofa:
[MSDN crypto docs state the opposite](https://msdn.microsoft.com/en-us/library/ie/dn302339%28v=vs.85%29.aspx), that IE11 `msCrypto` support is removed, replaced by  [the standard] `crypto` property.  Given that, is this diff needed?

broofa:
[Hmm.](https://github.com/bitwiseshiftleft/sjcl/issues/112)  :)

uuid.js
```diff
@@ -11,6 +11,9 @@
   // returns 128-bits of randomness, since that's what's usually required
   var _rng;
 
+  // Allow for MSIE11 msCrypto
+  var _crypto = _global.crypto || _global.msCrypto;
```
broofa: Can you reverse the order here?   I see some reports that IE11 defines `crypto`, which would render the above moot.  I.e., instead do this:
  
    var _crypto = _global.msCrypto || _global.crypto;

danorton2: Sure, but in the (very unlikely) case of both, or where someone else has defined crypto globally, do you prefer the non-standard name over the standard name?

I can say without any question that MSIE11 does not support "crypto" natively, but only supports msCrypto. (crypto was in MSIE10.)

danorton2: The MDN page you referenced recommends the order I proposed, as does this MS page:
https://msdn.microsoft.com/en-us/library/ie/dn265046%28v=vs.85%29.aspx

    var crypto = window.crypto || window.msCrypto;

broofa:
Thanks!

uuid.js
IanYates: If it helps, here's a screenshot from IE 11's console on a page in edge mode.
Note the different casing - Crypto vs crypto.  Weird / unexpected IMHO.

capture

danorton2: Crypto != crypto
msCrypto == crypto (for our purposes, here)

PR #114, opened 2015-06-12
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "node-uuid",
-  "version": "1.4.2",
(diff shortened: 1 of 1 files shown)


shinnn:
+1

coolaj86:
I'm using [mversion](https://github.com/mikaelbr/mversion), which git tags, bumps versions across {package,browser,component}.json, etc in sync.

Issue #119, opened 2015-08-19
babel throwing error on line 19


reharik:
Hi, I have a module that uses node-uuid 1.4.3.  I'm compiling said module with babel 5.8.2 and the flag ignore: false.  The ignore flag has babel compile your dependencies or rather setting it to false does.  The reason I have this set to false is because I have modules of my own that are using some es6 stuff, they are required via github and so to run the top level module I need to compile my dependent modules too.
Unfortunately this is causing an error that looks like this:
```
node_modules/node-uuid/uuid.js:19
  if (typeof _global.require == 'function') {
                    ^
TypeError: Cannot read property 'require' of undefined
```

I'm not sure what's going on but if some one does it would be great if you could let me know.
Thanks,
Raif

sergesemashko:
@reharik, did you find the solution? I have exactly the same error.
Thanks

reharik:
yes and no, I found a work around and that is you can give babel a ignore list.  I'm just having to put all stuff that's causing babel to blow up in the ignore list.  There are a couple of things.

sergesemashko:
just tried that and it works for me too. thanks

coolaj86:
This should no longer require an ignore list.

PR #121, opened 2015-08-31
Fallback to global for _global

johnhenry:
In some contexts, "this" may be undefined.
Falling back to global avoids this error.
http://stackoverflow.com/questions/9822561/why-is-this-in-an-anonymous-function-undefined-when-using-strict

### DIFF
--- a/uuid.js
+++ b/uuid.js
@@ -4,7 +4,7 @@
 //     MIT License - http://opensource.org/licenses/mit-license.php
 
 (function() {
-  var _global = this;
+  var _global = this || global;
 
   // Unique ID creation requires a high quality random # generator.  We feature
   // detect to determine the best RNG source, normalizing to a function that


biochimia:
FWIW, this fixes an issue with using `node-uuid` with babeljs:

```
node-uuid/uuid.js:19
  if (typeof _global.require == 'function') {
                    ^

TypeError: Cannot read property 'require' of undefined
```

biochimia:
The issue I mention was previously reported here as #119. The same (or similar) issues have been reported in packages depending on node-uuid: request/request#1661, and HewlettPackard/grommet#179, the latter fixed by dropping the `node-uuid` dependency.

mmahalwy:
Need this!

pollen8:
same issue here with babel, I can confirm that the PR resolves the issue for me.

mmahalwy:
Let's get it merged?

coolaj86:
Check out v1.4.6 and let me know if the issue is resolved.

PR #135, opened 2016-02-09
Update uuid.js

### LINKED ISSUE #134: Is there a way to setup node environment when window global is present?
Hi, so you are deciding between `browser` or `node` environment based on presence of window variable:

``` js
if (_window) {
  setupBrowser();
} else {
  setupNode();
}
```

where `_window` is

``` js
'undefined' !== typeof window ? window : null
```

We have a `window` global in node though and we would like to have the `node` part with `crypto.randomBytes` available. So we get rid of:

```
[SECURITY] node-uuid: crypto not usable, falling back to insecure Math.random()
```

Is there any workaround for this? 
Or are you willing to change the "decision" part for node to be more complex than just `window` being undefined? (e.g. having a `window.iAmStillNode` set to `true` or something more complex)

tracker1:
fixes #134

### DIFF
--- a/uuid.js
+++ b/uuid.js
@@ -269,4 +269,4 @@
 
     _window.uuid = uuid;
   }
-})('undefined' !== typeof window ? window : null);
+})('undefined' !== typeof window && window.self === window ? window : null);


amannn:
Any updates on if this is going to be merged?

jakubzitny:
What does @broofa think?

amannn:
I just did a check, this reports a jsdom environment as a browser which leads to printing the false warning in a test runner like Jest.

I've always used [detect-node](https://github.com/iliakan/detect-node/blob/master/index.js) for these tests which always worked pretty well.

What do you think @broofa ?

jakubzitny:
@amannn I've also realized this a while ago, totally forgot about this MR. Agreed though, `detect-node` or similar would be much better.

BTW @broofa, there is not a lot of maintaining activity on this project recently and that's pity because it's quite popular. Have you considered adding new maintainers to this project or maybe transfering the maintenance to another namespace (e.g. a company with OSS activities actively using this)?

(cc @coolaj86)

masylum:
It would be great to get this merged. This makes testing extremely noisy

PR #143, opened 2016-08-05
Whitelist package files

dguettler:
As part of ember-cli/ember-cli#6149, this PR aims to reduce the number of files downloaded for this package.


defunctzombie:
meh

PR #144, opened 2016-08-22
vars to const

capaj:
most people are writing for node.js 4 and up. It makes more sense to have examples written for the majority.


defunctzombie:
readme changed, will fix separately

PR #156, opened 2016-11-23
Correct bold

kuba-orlik:
I think you meant that part to be displayed as bold :)


defunctzombie:
Thanks!

Future note, helps if you are more descriptive in the commit message (i.e. correct markdown on readme, or README: fix markdown) so the commit history is clearer to those reading it.

Issue #154, opened 2016-11-23
Proposal: Separate submodules for different UUID versions


broofa:
I expect it's rare for someone using this library to want both v1() *and* v4() support.  99% of the time they'll only care about one flavor.  Thus, I think it would make sense to separate the code for the different versions into separate submodules.  So, for example ...

```javascript
var uuid = require('uuid/v1'); // returns v1() for timestamp-based UUIDs
var uuid = require('uuid/v4'); // returns v4() for random UUIDs
```

One immediate benefit is that if you don't use v1(), you don't need to deploy the v1() code, which is ~75% of the code in this module.

Another benefit is that it becomes straight-forward to add support for v3() and v5() UUIDs, which has occasionally been requested, but that I resisted in the original `node-uuid` library because it required shipping JS code to do MD5 / SHA hashes.  But now that `uuid` puts the burden for packaging on the user, it'd be straight-forward to implement those versions as well (and provide the requisite md5.js/md5-browser.js or sha.js/sha-browser.js modules as needed for node/browser support).

Thoughts?  Is this a reasonable direction to take this library?

defunctzombie:
👍 from me. No maintenance burden on our end to split the code out into the separate files.

Eventually this won't even be needed with the various "tree shaking" and optimization features going into rollup and webpack bundlers. They will be able to remove functions from the bundle which are not used.

reintroducing:
I just ran a yarn install with uuid 3.0 and am using the code from the readme:

```
import {v4} from 'uuid/v4';
...
// later using v4() to generate the uuid
```

Seeing this in terminal:
ERROR in ./core/js/ui/TextInput.jsx
Module not found: Error: Cannot resolve module 'uuid/v4' in /Users/przybylski/retrofit/dev/code/front-end/core/js/ui
 @ ./core/js/ui/TextInput.jsx 19:9-27

Any idea why this would be happening?

reintroducing:
Hmm, looks like yarn is still pulling the version that has uuid.js in the code hence why this isn't working. Any way you can publish a new version to npm to fix this?

broofa:
@defunctzombie - based on your comments on #159, I'll let you publish this.

defunctzombie:
Released in v3.0.1

ganqqwerty:
@defunctzombie  Does it mean that the uuid library is [tree-shakeable](https://webpack.js.org/guides/tree-shaking/)? Do you need to declare anything within the `sideEffects` section?

PR #159, opened 2016-11-24
uuid versions into separate modules, per #154

broofa:


component.json
```diff
@@ -1,6 +1,6 @@
 {
```
defunctzombie: We should just drop this file. I forgot to delete it when I merged the two repos.

package.json
```diff
@@ -1,14 +1,13 @@
 {
   "name": "uuid",
-  "version": "3.0.0",
+  "version": "3.1.0",
```
defunctzombie: no version changes in PRs, those happen after once ready to release (with changelog updates accordingly).

lib/rng-browser.js
```diff
@@ -1,32 +1,37 @@
...
       if ((i & 0x03) === 0) r = Math.random() * 0x100000000;
-      _rnds[i] = r >>> ((i & 0x03) << 3) & 0xff;
+      rnds[i] = r >>> ((i & 0x03) << 3) & 0xff;
     }
 
-    return _rnds;
+    return rnds;
   };
+
+  rng.isCrypto = false;
```
defunctzombie: unrelated to separating out into files?

defunctzombie:
👍 overall, just need to remove the version bump (FYI use the defunctzombie/changelog module to manage the changelog and release versions separately from PRs once I know what will go into the version). Also the addition of the new isCrypto api should be a separate PR.

PR #161, opened 2016-11-25
delete component.json

defunctzombie:


broofa:
In case anyone asks why this went away: https://github.com/componentjs/component/issues/639

PR #175, opened 2017-01-30
Is there a value to have an output in base64 encoding?

wood1986:
If so, I will write the test


broofa:
1. Not part of RFC spec
2. Doesn't work in browser
3. Unrelated changes that break things (testing for falsy != testing for undefined)
4. For the small minority of users that need this, this can easily be done outside via a helper function outside of this module.

PR #179, opened 2017-02-17
Add "uuid.isValid(str)".

michael-ericson-tanium:
This PR adds `uuid.isValid(str)` to test if the argument contains a valid UUID string.


broofa:
This has been discussed before. See #72.  

FYI, a more rfc-compliant regex: https://github.com/kelektiv/node-uuid/pull/72#r6502478

PR #182, opened 2017-02-28
TypeScript support

gautamsi:
Not sure if you would be interested in TypeScript version. It can be helpful in some cases.


broofa:
Not interested, but thanks for offering.

PR #183, opened 2017-03-01
(fix) Add .npmignore file to exclude test/ and other non-essentials

### LINKED ISSUE #181: Tests included in published package
```sh
$ npm install uuid
/temp
`-- uuid@3.0.1 
$ ls node_modules/uuid/test
mocha.opts test.js
```
See [Keeping files out of your package](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package) and [files](https://docs.npmjs.com/files/package.json#files) section in package.json.

mariotacke:
This fixes #181 

Personally, I would go one step further and also remove `HISTORY.md` and `AUTHORS` from the packed module. This is obviously up to the owner. I can adjust the PR to include those should that be desired.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,5 @@
+test/
+
+.gitignore
+.travis.yml
+.npmignore


broofa:
Looks reasonable to me.

mariotacke:
I added `AUTHORS` to `.npmignore` as well. Looks like we cannot ignore `HISTORY.md` as it is packed by default.

defunctzombie:
The ignoring of authors is superfluous and may detract from tracking down authors without having to look up the information from other sources.

mariotacke:
I removed that commit per your suggestion.

fanatid:
why not `files` in `package.json`? white list is better than ignore

mariotacke:
@fanatid , makes sense. Let's determine which files should be included. IMO only the code files should be included per package.json.

fanatid:
just for note, some files will be always included/ignored: https://github.com/npm/npm/blob/v4.5.0/lib/utils/tar.js#L66

PR #184, opened 2017-03-06
Fix use of `global` in browser environment

Omniroot:
This PR fixes errors that occur in browser environments due to the `global` variable not being present. Still ensures backwards compatibility for node environments or others using global.

### DIFF
--- a/lib/rng-browser.js
+++ b/lib/rng-browser.js
@@ -4,7 +4,9 @@
 // feature-detection
 var rng;
 
-var crypto = global.crypto || global.msCrypto; // for IE 11
+var _global = global || window;
+
+var crypto = _global.crypto || _global.msCrypto; // for IE 11
 if (crypto && crypto.getRandomValues) {
   // WHATWG crypto RNG - http://wiki.whatwg.org/wiki/Crypto
   var rnds8 = new Uint8Array(16);


broofa:
@defunctzombie, what's your package.json browser field spec say about how the `global` variable is defined in browser environments?  Is this supposed to be handled by the packaging utility?

Omniroot:
@broofa I'm not sure I understand. The `browser` field simply tells the bundler which files to use if it's compiling for a browser environment. Most browsers, to my knowledge, don't have a `global` variable.

broofa:
Most browsers don't have `require()`, either.  That's what packagers are for - they supply the environment needed to load CommonJS-compliant code.  It's just a question of whether or not the packager should be responsible for mapping `window` to `global`.

defunctzombie:
@broofa it does not speak about the global; this is left to the packaging utility but generally implied that the packaging utility will support it as they are effectively the same.

broofa:
@Omniroot What packaging tool are you using?  I think we'd prefer to see this issue fixed in that tool (so this issue is fixed for all projects) rather than hacking in a fix here.

Omniroot:
@broofa, it seems something was wrong with our webpack configuration, although I can't seem to reproduce the issue. I didn't realize that webpack handles the use of `global`, but I guess that makes sense given its purpose.

Omniroot:
For anyone looking at this in the future, it turned out to be a problem with the `babel-loader` configuration when using webpack 1. The babel presets needed to be defined in a `.babelrc` file instead of in the `query` parameters of the loader. I'm not sure why that would make this big of a difference, but it seems that was the problem as it is working fine now.

PR #188, opened 2017-04-20
V5 support

broofa:
Here's the v5 code I've been noodling on.

I'm comfortable with the API, and it works in node (via `npm test`), but I haven't done any in-browser testing.  There's also room for performance improvements, as currently I convert the namespace strings to byte arrays on every call, which is less than ideal, but I have limited time to work on this.  I think we can deal with that later, if it ever becomes an issue people care about.

### DIFF
--- a/lib/sha1.js
+++ b/lib/sha1.js
@@ -3,8 +3,19 @@
 var crypto = require('crypto');
 
 function sha1(bytes) {
-  if (Array.isArray(bytes)) bytes = Buffer.from(bytes);
-  return crypto.createHash('sha1').update(bytes).digest();
+	// support modern Buffer API
+	if (typeof Buffer.from === 'function') {
+		if (Array.isArray(bytes)) bytes = Buffer.from(bytes);
+		else if (typeof bytes === 'string') bytes = Buffer.from(bytes, 'utf8');
+	}
+
+	// support pre-v4 Buffer API
+	else {
+		if (Array.isArray(bytes)) bytes = new Buffer(bytes);
+		else if (typeof bytes === 'string') bytes = new Buffer(bytes, 'utf8');
+	}
+
+	return crypto.createHash('sha1').update(bytes).digest();
 }
 
 module.exports = sha1;
--- a/lib/sha1.js
+++ b/lib/sha1.js
@@ -1,0 +1,10 @@
+'use strict';
+
+var crypto = require('crypto');
+
+function sha1(bytes) {
+  if (Array.isArray(bytes)) bytes = Buffer.from(bytes);
+  return crypto.createHash('sha1').update(bytes).digest();
+}
+
+module.exports = sha1;
--- a/v5.js
+++ b/v5.js
@@ -1,0 +1,40 @@
+var sha1 = require('./lib/sha1-browser');
+var bytesToUuid = require('./lib/bytesToUuid');
+
+function uuidToBytes(uuid) {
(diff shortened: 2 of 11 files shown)


broofa:
Hmm... failing in node 0.12 and 0.4.   Someone want to take a look at what's going on here? (I don't have those node versions installed, and don't have time to dive into this atm.)

mike-marcacci:
@broofa this looks really great – thanks for adding this so quickly!! I ran these tests on node v4 in a docker environment, and didn't have any issues... I also compared the crypto and buffer docs between 4 and 7, and nothing has changed that should effect these. Kind of mysterious... I'm going to try again with 0.12

mike-marcacci:
Ah, never mind – found it.

lib/sha1.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+var crypto = require('crypto');
+
+function sha1(bytes) {
+  if (Array.isArray(bytes)) bytes = Buffer.from(bytes);
```
mike-marcacci: You'll want to explicitly create a buffer from a string, so you can specify the encoding (see PR #190).

mike-marcacci:
Anything else I can do to help move this?

broofa:
@defunctzombie I believe we're ready to roll this out.  Suggest we roll as v3.1.  Thoughts?

lib/sha1.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+var crypto = require('crypto');
+
+function sha1(bytes) {
+	// support modern Buffer API
+	if (typeof Buffer.from === 'function') {
+		if (Array.isArray(bytes)) bytes = Buffer.from(bytes);
+		else if (typeof bytes === 'string') bytes = Buffer.from(bytes, 'utf8');
```
defunctzombie: nitpick: prefer `{ }` even for single line if statements

defunctzombie:
I don't know what the reasoning behind v5 is vs using v4 but overall this looks fine. There will be a bit of size growth in the browser version due to the addition of v5 but I suppose those that really care can import/require the specific version file (probably won't happen in practice tho).

defunctzombie:
I think 4.0 is probably better choice given the change package size but 3.1 is technically ok because no existing APIs have changed.

mike-marcacci:
+1 on 3.1 just out of semver OCD.

@defunctzombie the rationale for v5 is the creation of deterministic UUIDs, which is unbelievably useful when working with distributed systems.

broofa:
@defunctzombie Actually this won't increase the browser package size since `index.js` doesn't have a v5 dependency.  v5 is only available via `require('uuid/v5')`.

broofa:
... note, too that bumping to v4.0.0 and removing index.js wouldn't be the worst idea.  The only issue with that is that there would no longer be an entry point for creating a pre-built, standalone file.  But... we're already drinking the module packager cool-aid here, so probably not that big a deal.

v5.js
```diff
@@ -0,0 +1,40 @@
...
+    bytes.push(parseInt(hex, 16));
+  });
+
+  return bytes;
+}
+
+function stringToBytes(str) {
+  str = unescape(encodeURIComponent(str)); // UTF8 escape
+  var bytes = new Array(str.length);
+  for (var i = 0; i < str.length; i++) bytes[i] = str.charCodeAt(i);
```
defunctzombie: nit: prefer { } even for single line blocks.

PR #191, opened 2017-05-05
refactor formatting and expose it via the API

andyburke:
I was looking for a way to format a uuid that I had generated in a buffer, but there was no formatting exposed via the API. I took a stab at it and am interested in feedback to see if this is something you would consider bringing into the mainline.


broofa:
The RFC is [very specific](https://tools.ietf.org/html/rfc4122#page-4) about the string representation of UUIDs.  That's a Grand and Glorious thing.  I see no benefit in supporting non-compliant formats.

andyburke:
@broofa I was unaware of the RFC around the formatting. I've removed the configuration of the format. Still, this PR would add an exposed formatter (which is what I originally needed, just a way to format a UUID into the proper format).

I'm willing to work with you on other issues you might have with this to try to move forward.

PR #193, opened 2017-05-16
fixup README (and publish?)

broofa:
@defunctzombie Let's get this published (as v3.1).  What else is needed for that?


broofa:
See #198

PR #196, opened 2017-06-02
Add v5 to cli

styfle:
The CLI script in `/bin` was missing the the v5 arg. This adds the argument.

### DIFF
--- a/bin/uuid
+++ b/bin/uuid
@@ -17,8 +17,8 @@ if (null == arg) {
   process.exit(0);
 }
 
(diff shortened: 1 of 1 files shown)


broofa:
`uuid v5` throws right now, because uuid/index.js doesn't actually export a `v5` property.    Also, v5 requires `name` and `namespace`.

styfle:
Should v5 be accessible via CLI?

broofa:
Moving to #197

PR #199, opened 2017-06-13
Expose a uuid formatter in the API

andyburke:
This change exposes a formatter through the API. The formatter produces RFC-compliant UUIDs without any configurability. This is useful for anyone who has UUIDs in a buffer but wants to be able to format them according to the RFC.

This is related to #191 (have not heard back on queries there, also updated with latest changes in master).


broofa:
Two questions this raises:

1. Should this module expose a buffer/array -> uuid string API?

https://github.com/zefferus/uuid-parse exposes the parse/unparse methods that are internal to this module already.  Any reason that won't work for you?

2. What's the correct implementation for that API?

I don't know that the implementation matters all that much, as long as we agree on what the inputs and outputs should be.  That said, can you enlighten as to the ways in which your implementation is better?  Seems like the existing implementation is smaller, faster, and supports buffer offset.  :-/

broofa:
FWIW, the parse and unparse methods in uuid-parse used to be public methods on this module.  I removed them because they were a non-essential distraction from the core value of this module.

While it may seem trivial to just make the formatter method public, it implies a broader commitment to not just formatting, but parsing as well, and that leads to validation... it's a slippery slope.  I've found it simpler to just not open that can of worms.  I see little reason to reverse that decision, I'm afraid.

Feel free to argue to the contrary, but barring a convincing argument to the contrary, I'm inclined to close this.

PR #200, opened 2017-06-18
Update README

tomciopp:
Fix typo should reference v5 instead of v4


broofa:
good catch, thanks!

broofa:
FYI, #204 fixes this

broofa:
Fixed in #208

PR #201, opened 2017-06-21
Fix buffer not being modified (uuid v5)

Landmaster:
This pull request fixes the `buf` parameter being silently ignored by the code for generating a v5 uuid.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.sw*
+.idea/**
 .DS_Store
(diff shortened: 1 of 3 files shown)


broofa:
Looks good.  Can you add a unit test to catch this case so it doesn't happen again?  Thanks!

Landmaster:
@broofa Some unit tests have been added to test the functionality.

PR #205, opened 2017-06-24
please add missing values

tsypa:
missing pre-defined values for namespace IDs from RFC 4122 ( Appendix C)
https://tools.ietf.org/html/rfc4122#appendix-C

### DIFF
--- a/v5.js
+++ b/v5.js
@@ -46,5 +46,6 @@ function v5(name, namespace, buf, offset) {
 // Pre-defined namespaces, per Appendix C
 v5.DNS = '6ba7b810-9dad-11d1-80b4-00c04fd430c8';
 v5.URL = '6ba7b811-9dad-11d1-80b4-00c04fd430c8';
-
+v5.OID = '6ba7b812-9dad-11d1-80b4-00c04fd430c8';
+v5.X500 = '6ba7b814-9dad-11d1-80b4-00c04fd430c8';
 module.exports = v5;


broofa:
@tsypa - Why are you interested in these namespaces?  Are you submitting this PR simply for the sake of completeness with the spec, or because you have an actual need to generate IDs in the OID or X.500 namespaces?  If the latter, which namespace, and what might your use case be? (I'm genuinely curious.)

I ask because I deliberately chose to omit these definitions since they did not appear to be of widespread interest.  And in the rare case someone does need to generate uuids for these namespaces, it's trivial to define them outside this module.

E.g. The [ISO OID namespace](http://www.oid-info.com/cgi-bin/display?tree=0,1,2&see=all) appears to mostly be of interest to people within the standards body community.  While the work this community does is important to a lot of people, I don't know that generating uuids in this namespace is of interest to anyone outside this community.

E.g.  The [X.500 namespace](https://en.wikipedia.org/wiki/X.500) is [criticized](https://en.wikipedia.org/wiki/X.500#Criticism), and described as having been [killed by LDAP 10 years ago](http://www.networkworld.com/article/2287928/infrastructure-management/x-500-vs--ldap.html).

We can leave this PR open for now, but barring arguments to the contrary I'm inclined to just close it and leave things as they are.

tsypa:
No, I don't need to generate uuids within this namespaces for now, but someone (i think)
This package described as rfc-compliant, and this is just two constants.

broofa:
Cool, thanks.  I'm going to leave these out until there's a demonstrated need for them.  I'm finding that err'ing on the side of simpler, less-krufty APIs is a good thing these days. :-)

PR #211, opened 2017-07-11
Expose newly added V5 in global namespace

dmlukichev:


broofa:
Closing. index.js is a legacy API that should not be enhanced. New users should explicitly require the version of uuid they need, e.g.  require('uuid/v4')

Issue #214, opened 2017-07-19
typos in the readme on NPM


blord-fullscreen:
These are typos I am assuming? Note that the Github readme doesn't have the first typo (arrow) but does have the second. Confusing for a newcomer.

![this-aint-right](https://user-images.githubusercontent.com/27975206/28344787-bd4b1f26-6bda-11e7-9af3-a99082acd86b.png)

Thanks!

Landmaster:
FIxed in #215, can be closed

PR #215, opened 2017-08-11
Corrected version from v4->v5 in README_js.md

poldridge:
This was previously fixed in https://github.com/kelektiv/node-uuid/commit/880d24e35d2a393b115ce3687465866211ad0680 ,
but was regressed in README_js.md and hence in the newly generated README.md

Also reran `npm run-script prepare` to update README.md from the corrected version


poldridge:
Fixes #214

PR #216, opened 2017-08-19
v5 additions (default namespaces, buffer inputs, cli binary args)

tschiemer:
Hi, many thanks for the library. Here my changes:

1. Added the Object ID and X500 default namespaces as they are also defined in rfc4122, appendix c. There is a total of 4 namespaces listed, but only two were defined in the library - so why not add the other two?
2. uuid/v5 now also takes a (nodejs) `Buffer` object as input and converts it into an array as needed assuming the bytes to be unsigned chars
3. for the cli I added the option `--hex` (before the `name` argument) as to indicate that the name argument is actually a hexadecimal string that is to be read as binary data

Regarding 2: For systems/browsers not supporting a Buffer-object/class I put a safeguard, but I'm not 100% sure wether this works - maybe someone who knows better can confirm?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Version 1 (timestamp):
 
 ```javascript
 const uuidv1 = require('uuid/v1');
-uuidv1(); // ⇨ '985123a0-7e4f-11e7-9022-fb7190c856e4'
+uuidv1(); // ⇨ 'aee02740-8590-11e7-b662-938de05596cd'
 
 ```
 
@@ -30,7 +30,7 @@ Version 4 (random):
 
 ```javascript
 const uuidv4 = require('uuid/v4');
-uuidv4(); // ⇨ 'df7cca36-3d7a-40f4-8f06-ae03cc22f045'
+uuidv4(); // ⇨ 'f576bd76-e5eb-4924-981b-631f245cbac4'
 
 ```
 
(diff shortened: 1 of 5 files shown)


broofa:
Hi, thanks for the PR!

Re: OID and X.500, please see https://github.com/kelektiv/node-uuid/pull/205

I'm open to adding 'Buffer' support to the v5 [and newly added v3] APis, but let's pull that out into a separate PR.  (And, for future reference, I believe lumping unrelated code changes together in a single PR is generally frowned upon.  Smaller, more focused PRs are easier to review and manage in my experience.)

For `--hex` option I'd like to hold off until we see more demand for this.  The CLI is a convenience for the simple use cases, and I don't see binary names as falling into that "simple" category.

Lastly, README me changes should be made to [README_js.md](../blob/master/README_js.md), which gets compiled into README.md when published.

I'm going to close this out.  Feel free to open another PR for the Buffer API changes.

Cheers!

tschiemer:
Hi!

Oh, thanks for pointing out my bad habit!

As for providing a use case of the `hex` option (and in principle the linked buffer input): I'm working with hardware components that occasionally provide IDs as arbitrary byte sequences which is not very string friendly. So being able to conveniently generate namespaced UUIDs from arbitrary `names` would be a helpful out-of-the-box feature.
But as you say, this might not be your typical use case.

I seem to have forgotten committing the REAME_js.md, sorry for that..

Disco!

PR #217, opened 2017-08-28
UUID v3 Support

abstein2:
Adds support for UUID v3 based on the code for v5. Since v3 and v5 are generated in almost exactly the same way the major addition to this is added `md5` (using crypto) and `md5-browser` (based on the implementation in https://github.com/blueimp/JavaScript-MD5) to the `lib` directory and then using them.

Updated tests, CLI commands, README_js.md, and tests. I needed v3 support for older systems as we migrate them to JavaScript.

Happy to make any additional changes.


broofa:
This looks good, thanks!  (It'll be nice to finally round out all the uuid versions in this module after all these years. :-D )

One change:  v3.js and v5.js are identical but for a [couple lines](https://github.com/kelektiv/node-uuid/blob/master/v5.js#L34-L35).  Let's factor that code into a common module (e.g. `lib/v35.js`?) thusly ...

v35.js:
```
module.exports = function(name, version, hash) {
  var uuid = function(name, version, namespace, buf, offset) {
    ... (guts of v3()/v5() implementations) ...
    var bytes = hash(namespace.concat(name));
    bytes[6] = (bytes[6] & 0x0f) | version;
    ...
  }
  uuid.name = name; // Set function name shown in debuggers
  return uuid;
}
```

v3.js
```
var v35 = require('./lib/v35.js');
var md5 = require('./lib/md5');
module.exports = v35('v3', 0x03, md5);
```

v5.js
```
var v35 = require('./lib/v35.js');
var md5 = require('./lib/sha1');
module.exports = v35('v5', 0x05, sha1);
```

abstein2:
Good idea. Updates made.

PR #219, opened 2017-09-07
eslint

broofa:
`eslint --fix` cleanup


lib/sha1.js
```diff
@@ -3,19 +3,17 @@
...
 
 function sha1(bytes) {
-	// support modern Buffer API
-	if (typeof Buffer.from === 'function') {
-		if (Array.isArray(bytes)) bytes = Buffer.from(bytes);
-		else if (typeof bytes === 'string') bytes = Buffer.from(bytes, 'utf8');
-	}
+  if (typeof Buffer.from === 'function') {
+    // Support modern Buffer API
+    if (Array.isArray(bytes)) bytes = Buffer.from(bytes);
```
defunctzombie: would be nice for eslint to require `{ }` I find this safe programming practice

PR #220, opened 2017-09-07
Defer random initialization of node and clockseq. Fixes #189

### LINKED ISSUE #189: Slow require() of v1
Using ```require("uuid/v1")``` under Electron, I've noticed anywhere from 2x to 4x increase in window load time. I've tracked this to the unwrapped call to ```rng()``` which in turn synchronously calls ```crypto.randomBytes()```. From the NodeJS docs:

 ```The crypto.randomBytes() method will block until there is sufficient entropy. This should normally never take longer than a few milliseconds. The only time when generating the random bytes may conceivably block for a longer period of time is right after boot, when the whole system is still low on entropy.```

This is unfortunately the behvaviour I'm seeing. The loading of the entire window is being held up by the call to ```crypto.randomBytes```. Changing my use of ```uuid/v1``` to ```uuid/v4``` fixes the problem.

If there is sufficient interest in modifying ```uuid/v1``` to use the asynchronous version of ```crypto.randomBytes``` or otherwise wrap the synchronous version so it is not called on require time I would be happy to open a PR.

broofa:


v1.js
```diff
@@ -6,17 +6,7 @@ var bytesToUuid = require('./lib/bytesToUuid');
...
-
-// Per 4.5, create and 48-bit node id, (47 random bits + multicast bit = 1)
-var _nodeId = [
-  _seedBytes[0] | 0x01,
-  _seedBytes[1], _seedBytes[2], _seedBytes[3], _seedBytes[4], _seedBytes[5]
-];
-
-// Per 4.2.2, randomize (14 bit) clockseq
-var _clockseq = (_seedBytes[6] << 8 | _seedBytes[7]) & 0x3fff;
+var _nodeId, _clockseq;
```
defunctzombie: prefer use of one line per var

PR #221, opened 2017-09-08
use typeof to check for crypto rather than global.  Fixes #185

### LINKED ISSUE #185: Provide instructions for usage with babel
There are issues with how the random number generator code is implemented. It fails with default settings in browserify and babel. (See #172)

There are a number of issue that describe the problem but none that detail the solution.

- There should be notes in the main readme for usage with popular packaging setups, like babel.
- The code should be patched so that it doesn't break by default.

The [offending line](https://github.com/kelektiv/node-uuid/blob/master/lib/rng-browser.js#L7) is relying on `global` being defined as `window`. For some reason `global === undefined` when this library is included in another project and compiled using babel/webpack.

broofa:


lib/rng-browser.js
```diff
@@ -2,25 +2,23 @@
 // browser this is a little complicated due to unknown quality of Math.random()
 // and inconsistent support for the `crypto` API.  We do the best we can via
 // feature-detection
-var rng;
 
-var crypto = global.crypto || global.msCrypto; // for IE 11
-if (crypto && crypto.getRandomValues) {
+if (typeof(crypto) != 'undefined' && crypto.getRandomValues) {
```
defunctzombie: `!==` ?

broofa: I'll address this and the other eslint issues you raised in the other PRs in the `eslint` PR.

Issue #173, opened 2017-01-08
Insecure random values should be opt-in


joepie91:
Right now, this module will automatically fall back to `Math.random()` in browser environments that don't support CSPRNG use, essentially failing open.

To operate securely, it should instead fail hard with an error, and only allow falling back to an insecure random source if the developer explicitly indicates that this is acceptable.

defunctzombie:
While I understand your point, I think the current behavior is more developer friendly and since this module isn't even aiming for strong crypto I feel less strongly that this is a real problem.

broofa:
I actually feel like this is a pretty legitimate request.  Most devs aren't aware of the security issues of less-than-crypto-quality PRNGs. Requiring them to opt-into the existing possibly-insecure behavior is probably a good thing.

... in the next major version, though.

Reopen?

defunctzombie:
Sure, if you want to reopen go for it.

My thinking was that even tho it isn't crypto quality, this lib isn't that hurt by not being crypto quality as fallback since it isn't a crypto library itself. But if you have a good idea for the API change we can do another major bump.

pgaubatz:
Hi @joepie91

You might have a look at my implementation that doesn't fallback to Math.random: https://github.com/pgaubatz/node-uuid

Cheers,
Patrick

broofa:
For the record, [crypto.getRandomValues is broadly supported](http://caniuse.com/#feat=getrandomvalues) these days.

willsr:
I think this a serious issue. We had issues with uuid collisions, specifically in test suites. The whole point of a UUID is to be universally unique...

broofa:
@willsr, can you elaborate?  UUID collisions are always a concern.  Specifically, what platform are you running on? What code are you running to generate UUIDs?  How often are you seeing collisions?  etc...

broofa:
For the record, the CVE for this issue: http://www-01.ibm.com/support/docview.wss?uid=swg21982849

SmaugTheGreat:
If you look at [this](https://medium.com/@betable/tifu-by-using-math-random-f1c308c4fd9d) article, you'll notice that it is a very severe problem as some browsers (most notably the google webcrawlers) fall back to a very weak implementation in Math.Random which guarantees collisions after just a few thousand invocations. This causes some severe problems in some applications and as such needs to be fixed. Snowplow has big problems because of this.

anagytherealone:
[...]

PR #225, opened 2017-10-01
Dubious rng

broofa:
#173 points out that quietly falling back to Math.random raises security concerns.  This diff causes `uuid` to throw an error if/when Math.random would be used to create UUIDs, provides a hook for suppressing that error, and links to a wiki page (https://goo.gl/vz3J87) that explains the error, the hook, and a possible workaround for generating high-quality UUIDs using SJCL crypto library.


broofa:
BTW, I'm thinking we bump the major version for this, since it may throw in cases where it didn't previously.

lib/rng-browser.js
```diff
@@ -2,30 +2,31 @@
...
+  return rnds8;
+}
 
-  module.exports = function mathRNG() {
-    for (var i = 0, r; i < 16; i++) {
-      if ((i & 0x03) === 0) r = Math.random() * 0x100000000;
-      rnds[i] = r >>> ((i & 0x03) << 3) & 0xff;
-    }
+function mathRNG() {
+  if (!rng._dubious) throw new Error('No good RNG available. See https://goo.gl/vz3J87');
```
defunctzombie: lint?

defunctzombie:
I would probably consider allowing the user to specify the crypto, or honestly, maybe recommend they polyfill the crypto functions the library needs if their browser does not have them? I do like the idea of a library that ships with the support it needs so another option is to not change anything and simply let the user polyfill the crypto function to throw if they don't want to suppose "weak" crypto browsers.

broofa:
> let the user polyfill the crypto function

Oh, I like that idea.  Can't really think of any downsides to this at the moment.

> another option is to not change anything and simply let the user polyfill the crypto function to throw

This is basically asking users to fix the problem themselves.  I don't think we can say we've fixed this bug if naive users are still vulnerable to the issue.  I think the spirit of this change is to complain loudly if there's an issue, and then inform the user about the problem and what options are available for fixing it.

broofa:
<rant>npm is filled with prng modules that are just wrappers around window.crypto or window.msCrypto.  Boo!</rant>

vvo:
I don't have the required knowledge to review this sorry, maybe @defunctzombie can do another pass?

broofa:
@vvo: No worries - I'm not sure we've resolved how best to address this.  As @defunctzombie says, asking the user to find a polyfill may be the best solution here, but it's up for debate.

broofa:
Apparently I don't have time/interest to push this through.

PR #227, opened 2017-10-09
Export v3 and v5 from main package

mike-marcacci:
On the server or when using es6 imports with tree shaking, it's idiomatic to use:

```es6
import { v4, v5 } from 'uuid';
```

instead of:

```es6
import v4 from 'uuid';
import v5 from 'uuid/v5';
```

This makes the former possible.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,12 @@
 var v1 = require('./v1');
+var v3 = require('./v3');
 var v4 = require('./v4');
+var v5 = require('./v5');
 
 var uuid = v4;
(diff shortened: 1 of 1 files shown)


broofa:
Hi Mike, thanks (again) for contributing.

I've deprecated the top-level API object because I don't think it adds much value for most users, and creates a cross-version dependency that in almost all cases pulls in extra code that won't get used.  It's going away in the next major rev (if/when) that happens.  (This is documented in the README.)  Thus, I'm not interested in adding to that API.

PR #232, opened 2017-11-08
why would one be failing and not the other?

wenzowski:
Confused as to why uuids are not generated when `uuid.v1()` or `uuid.v4()` is called from `Array.map()`.

failing & passing tests provided to isolate

```
node --version
v8.9.1
```

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -258,3 +258,17 @@ test('ids spanning 1ms boundary are 100ns apart', function() {
   var dt = parseInt(after, 16) - parseInt(before, 16);
   assert(dt === 1, 'Ids spanning 1ms boundary are 100ns apart');
 });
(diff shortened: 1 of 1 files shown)


broofa:
`Array#map()` passes args to the iterator function that are inconsistent with the args that `v1()` expects.  Specifically, the 2nd arg passed in is the Array index.  But `v1()` looks for a `buf` arg there.  If defined, `v1()` sets the elements of that array and returns `buf`... i.e. when invoked directly as an iterator, `v1()` will just return the index for all but the first (index == 0) invocation.

wenzowski:
Right. Silly me for not reading carefully! Yes, definitely an edge case. Thanks for the clarification :)

PR #233, opened 2017-11-09
Add missing exports

w8jcik:
Is there any reason v3 and v5 are not exported from index?


w8jcik:
I have found this explanation https://github.com/kelektiv/node-uuid/pull/227

PR #235, opened 2017-11-16
Auto generate CHANGELOG.md. Fix #229

### LINKED ISSUE #229: Add 3.1.0 to History
Release 3.1.0 is not documented in the History file. Could this be added?

broofa:
This replaces HISTORY.md with CHANGELOG.md, and autogenerates it from commit history when `npm run changelog` is run.

This is a bit of a change in that we (well, me, really) haven't done a good job of capturing the major changes in releases as github issues.  So the CHANGELOG items are substantively different from what's in HISTORY.md currently.  But... dunno... as bad as I am at creating issues, I'm even worse at manually maintaining HISTORY files.  I'd rather have an automated tool managing this stuff and just
 be better at filing issues to capture the actual changes taking place. 🤷‍♂️ 

Another thing that's a little weird about this is that you have to generate the git tag prior to running this command, which means the commit that updates CHANGELOG ends up *after* the tag.  I've worked around this in the past on other projects (namely node-mime) by using `rebase -i` to swap the order of the commits but... it's kind of clumsy.  I'm open to suggestions for better ways to manage this.

Thoughts?


simonfox:
I've used [standard-version](https://github.com/conventional-changelog/standard-version) successfully for a while now. We do use GitHub issues, but by following the commit conventions, you _could_ get away without those. No such issues with the tagged commit either.

simonfox:
Should probably add that we change our repo PR merge settings to only allow _Squash and Merge_. This way you can make sure that the commits that hit master follow the conventions required for standard-version.

I learnt about standard-version after contributing to [Aurelia](https://github.com/aurelia). Their contributing guidelines have a good summary of conventions that work with standard-version https://github.com/durandalproject/about/blob/master/CONTRIBUTING.md#commit

broofa:
@simonfox Cool, I'll look into it. Thanks!

broofa:
Re-submitting as #246 (Use `standard-version`, per @simonfox's suggestion)

PR #236, opened 2017-11-22
Adds the Nil UUID as a constant to the module exports.

michaelshaffer37:
This seems kinda trivial, but I had an occasion a few weeks ago where I needed the [NIL UUID value](https://tools.ietf.org/html/rfc4122#section-4.1.7) obviously this is easy enough to just define as a constant.  But I thought that it would have been included as part of this package and was surprised not to find it as it's part of the UUID specification.  However when I went to add it to node-uuid I initially thought to add it in the same way that the v5 includes the DNS and URL namespaces.  However the NIL uuid is categorically different and should really be included with all versions 1, 3, 4 & 5.
I considered centralizing the definition but the overhead of a require just for a static constant seemed a bit much.
If this implementation is alright, I'd like to update the docs with this PR.

Please let me know what you think.

### DIFF
--- a/lib/v35.js
+++ b/lib/v35.js
@@ -48,6 +48,7 @@ module.exports = function(name, version, hashfunc) {
   // Pre-defined namespaces, per Appendix C
   generateUUID.DNS = '6ba7b810-9dad-11d1-80b4-00c04fd430c8';
   generateUUID.URL = '6ba7b811-9dad-11d1-80b4-00c04fd430c8';
+  generateUUID.NIL = '00000000-0000-0000-0000-000000000000';
 
   return generateUUID;
 };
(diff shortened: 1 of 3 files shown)


broofa:
Thanks for the suggestion.   While I understand the case you're making, I don't think there's enough value in this to justify inclusion in this module.   It's akin to the OID and X.500 namespace uuids...  technically these are part of the spec but, in practice, they're very rarely used and, as such, not worth cluttering up the API.  Especially since they're trivially easy to define ad-hoc.

I'll even go so far as to suggest that the Nil UUID shouldn't be part of the spec, as it is redundant with (and inferior to) the native `nil`/`null`/`None`/`NULL` types that nearly all modern languages and data stores support.

PR #240, opened 2017-11-28
Publish v3.2.0

broofa:
This is to push v3 support out.
I'll tag and update CHANGELOG.md once this is merged, and then publish to npm.


broofa:
FYI, 'Need to fixup #241 before I roll this out.

PR #249, opened 2018-01-03
Remove Date as an available type for msecs option of v1

### LINKED ISSUE #248: README incorrectly defines `msecs`
The README docs for the `v1()` options states:

> `msecs` - (Number | **Date**) Time in milliseconds since unix Epoch. Default: The current time is used.

But `Date` instances do not work as expected. Only millisecond values do:

```js
> uuid.v1({ msecs: new Date() })
'00000000-0000-1000-b5d5-0d7fa4a8285e'

> uuid.v1({ msecs: new Date().getTime() })
'3a7ea410-effa-11e7-b5d5-0d7fa4a8285e'
```

I'm happy to submit a PR to fix this, but is this something where you would like the README to be corrected, or would you like the `v1()` implementation to accept `Date` instances for the `msecs` option?

rmm5t:
Fixes #248 

I figured this was the simplest approach and the most self-documenting considering that the option is named `msecs`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Generate and return a RFC4122 v1 (timestamp-based) UUID.
 
   * `node` - (Array) Node id as Array of 6 bytes (per 4.1.6). Default: Randomly generated ID.  See note 1.
   * `clockseq` - (Number between 0 - 0x3fff) RFC clock sequence.  Default: An internally maintained clockseq is used.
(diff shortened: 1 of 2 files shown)


rmm5t:
closing then re-opening to trigger a new build, because Travis went on the fritz during the last build.

broofa:
Sorry, should have commented here earlier.  The RunMD step needed some love so I've reworked this as #251

PR #251, opened 2018-01-04
Fix #248

### LINKED ISSUE #248: README incorrectly defines `msecs`

broofa:
Also updated to use `runmd`@v1.0.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+<!--
+  -- This file is auto-generated from README_js.md. Changes should be made there.
+  -->
 
 # uuid [![Build Status](https://secure.travis-ci.org/kelektiv/node-uuid.svg?branch=master)](http://travis-ci.org/kelektiv/node-uuid) #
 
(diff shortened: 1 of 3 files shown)


rmm5t:
Strange that you ignored #249, but `¯\_(ツ)_/¯`

broofa:
Yeah, sorry, I should have commented here.  I appreciate the contribution / would have merged it, but 'turned out the RunMD stuff was broken and needed some work so... yeah. `¯\_(ツ)_/¯`

PR #254, opened 2018-01-13
Added build infrastracture, building for 2 targets - cjs and es

Andarist:
This allows better integrations with browser bundlers like webpack.

Thanks for distributing also "module" builds it can be tree-shaken from consumers' builds. Also other optimization techniques are possible thanks to that - such as scope hoisting.

If you have any questions regarding changes being made, please just ask.


broofa:
'Lot of code churn.  Tests don't pass.  Introduces [non-trivial](http://npm.broofa.com/?q=babel-cli) dev [dependencies](http://npm.broofa.com/?q=babel-plugin-transform-es2015-modules-commonjs).  Unclear as to what actual problem(s) this is solving.

I guess the main question here is, "does this actually improve anything?"  Tree shaking does not add value.  As long as users are requiring the specific version they want, there's not going to be any dead code to remove.

I know what var hoisting is, but I'm unclear what you mean by scope-hoisting in this context, or how it adds value.

Andarist:
> 'Lot of code churn.

Could u elaborate?

> Tests don't pass.

Fixed, sorry for this - thought I've fixed all of them, but it seems I've fixed some other tests and thought I'm done.

> Introduces non-trivial dev dependencies. Unclear as to what actual problem(s) this is solving.
> I guess the main question here is, "does this actually improve anything?" Tree shaking does not add value. As long as users are requiring the specific version they want (e.g. require('uuid/v4'), there's not going to be any dead code to remove.)

That's true for "direct" consumers, but if you consider consumer of packages relaying on your module the case gets more complicated. Various tools under various circumstances might bail out from removing `uuid` if it got requested by otherwise tree-shakeable module.

> I know what var hoisting is, but I'm unclear what you mean by scope-hoisting in this context, or how it adds value.

Scope hoisting is bundlers' technique to put most modules in the same scope (top level, let's say). Historically i.e. webpack has used runtime closure wrappers to isolate each module, but that adds to runtime (especially startup) costs. `rollup` has popularized the idea of concatenating modules into a single scope (after renaming identifiers, and possibly doing some tricks for cyclic dependencies and other edge cases) - `webpack` is catching on and has introduced a plugin for doing this in webpack3 which will be enabled by default in webpack4 in production mode. 

Unfortunately `webpack` just bails out from both tree-shaking and scope hoisting optimizations as soon as it sees non-harmony module.

Andarist:
Note: to support node 0.12 I had to do some hoops with:
```js
  if (Object.getOwnPropertyDescriptor(generateUUID, 'name').configurable) {
    Object.defineProperty(generateUUID, 'name', { value: 'name' });
  } else {
    generateUUID.name = name;
  }
```
and had to disable strict mode. If you'd ask me, I'd just release major release with dropped 0.12 support - but that's ofc your decision.

We could also play a little bit with babel setup to disable strict mode only for that single file, but that would complicate things a little bit, rather not worth it.

package.json
```diff
@@ -11,20 +11,29 @@
...
     "md": "runmd --watch --output=README.md README_js.md",
-    "prepare": "runmd --output=README.md README_js.md"
+    "prepare": "npm run build && runmd --output=README.md README_js.md"
   },
   "browser": {
     "./lib/rng.js": "./lib/rng-browser.js",
     "./lib/sha1.js": "./lib/sha1-browser.js",
-    "./lib/md5.js": "./lib/md5-browser.js"
+    "./lib/md5.js": "./lib/md5-browser.js",
+    "./src/rng.js": "./src/rng-browser.js",
```
vvo: duplicate?

vvo:
There are many code that is not tied to this current change as stated in the PR (readme etc..), can you clean maybe?

package.json
Andarist: Nope, this is intentional - i rewrite here imports for both types of modules. If you use node or "cjs bundler" you will end up with files from `lib`, if you use module aware bundler like webpack2+ you will end up with files from `src`

Andarist:
Actually I think the only unrelated files are those READMEs. They got on my machine thanks to `prepare` script, then I've seen that the generation script was broken, so I've fixed it along the way to confirm my changes and stuff. Personally I dont consider it being an issue, especially that it is a fix (only not exactly related one). If you feel otherwise I can revert those changes though

package.json
Andarist: Those are not duplicates, there are 3 "rewrite" keys from lib to lib directory and 3 "rewrite" keys from src to src. This package has **legacy** `index.js` - it actually has no main/module fields, because you are supposed to "cherry-pick" what you want by requiring/importing i.e. `uuid/v4`

The resolution algorithms works like this:
- `uuid` directory is found
- `v4` directory is found
- as this directory has only `package.json` (since this PR) with `main` and `module` fields they are followed regarding regular rules, node will follow `main`, webpack etc will follow `module`
- appropriate directory is visited (let's say we use webpack so its `src`) and for each requested file it is checked in "closest" `package.json` (closest from `uuid/src` is the one in the root of the package) if `browser` fields has a "rewrite" from the requested relative path to some other relative path - so webpack will follow this rewrite and use appropriate file which targets browser AND is written using es6 modules syntax

broofa:
FYI, I've merged some of the other PRs, which caused a few conflicts here.  My apologies.  But when you merge, this should get rid of the README & runmd diffs.

broofa:
Continuing the conversation from before ...

> That's true for "direct" consumers, but if you consider consumer of packages relaying on your module the case gets more complicated.

Depending on the top-level API has been explicitly deprecated for precisely this reason.  As I said, tree-shaking is of minimal value.

> Scope hoisting is bundlers' technique to put most modules in the same scope (top level, let's say)

Ah, thank you for the explanation.  This is an interesting new feature of packagers I was not aware of.  After doing a little reading [here](https://medium.com/webpack/brief-introduction-to-scope-hoisting-in-webpack-8435084c171f) and [here](https://docs.google.com/document/d/1E2w0UQ4RhId5cMYsDcdcNwsgL0gP_S6SDv27yi1mCEY/edit) and [here](https://github.com/webpack/webpack/issues/2873), I'm tentatively of the opinion that this isn't something worth being too concerned about.

For example, that latter link - the github issue on webpack - suggests that the perf gain here would  be about 0.05 msecs (250ms/5000 modules).

While I like the idea of moving to ES6 module notation, I'm a little concerned that this may break legacy support for the browser-ready versions.  E.g. what does wzrd.in do with ES6-style modules?

lib/rng-browser.js
```diff
@@ -3,29 +3,34 @@
 // and inconsistent support for the `crypto` API.  We do the best we can via
 // feature-detection
 
-var getRandomValues = (typeof(crypto) != 'undefined' && crypto.getRandomValues) ||
-                      (typeof(msCrypto) != 'undefined' && msCrypto.getRandomValues);
-if (getRandomValues) {
-  // WHATWG crypto RNG - http://wiki.whatwg.org/wiki/Crypto
-  var rnds8 = new Uint8Array(16); // eslint-disable-line no-undef
+var rng = /*#__PURE__*/function () {
+  var getRandomValues = typeof crypto != 'undefined' && crypto.getRandomValues || typeof msCrypto != 'undefined' && msCrypto.getRandomValues;
```
beck: This is missing binding: https://github.com/kelektiv/node-uuid/pull/257

beck:
Without tests, the refactoring here is hazardardous. For example, this knocks out a merged bugfix.

Some build infrastructure would be nice but it should be done without editing any of the lib's source. A good first step would be knocking out use of `wzrd.in`. As-is wzrd.in makes it difficult to test changes without first publishing.

Maybe something like:
```
# these bundles are git ignored
$(npm run)/browserify --standalone uuid index.js > index.bundle.js
$(npm run)/browserify --standalone uuidv1 v1.js > uuidv1.js
$(npm run)/browserify --standalone uuidv3 v3.js > uuidv3.js
....
```

☝️  would have made adding browser testing easier. Would also make it easy to add `prepublish` minification.

Then @broofa could push to https://cdnjs.com/libraries/node-uuid and the docs look like:

```diff
- <script src="http://wzrd.in/standalone/uuid%2Fv4@latest"></script>
+ <script src="https://cdnjs.cloudflare.com/ajax/libs/node-uuid/3.2.1/uuidv1.js"></script>
```

This is nice because the API stays the same and now the file name matches the function that gets attached to the window. Those who are using `wzrd.in` won't see any breakage.

After tests lands, rewrites and syntax modernization can be done with confidence.

lib/rng-browser.js
Andarist: well, it wasnt there when ive created/rebased this PR :P need to rebase again, gonna do that in a moment

Andarist:
> Without tests, the refactoring here is hazardardous. For example, this knocks out a merged bugfix.

This doesnt knock anything out... it just predates other PR, so naturally it has conflicts, but those will be fixed in a moment.

> Some build infrastructure would be nice but it should be done without editing any of the lib's source. A good first step would be knocking out use of wzrd.in. As-is wzrd.in makes it difficult to test changes without first publishing.

I will certainly give a thought how this can be tested, thanks for suggestions.

broofa:
Closing.   I understand why you want this, but this doesn't solve any real, tangible, user pain.  Thus, the increased build and code complexity isn't worth it.  I do appreciate the contribution, though.  'Might resurrect this in the future if any of the issues you've mentioned become significant enough to warrant addressing.

BenjaminVanRyseghem:
When rollup tries to convert this library, it ends up converting

https://github.com/kelektiv/node-uuid/blob/master/lib/v35.js#L47-L50
```js
  try {
    generateUUID.name = name;
  } catch (err) {
}
```

to 

```js
generateUUID.name = name;
```

which is broken.

My 2cts

PR #260, opened 2018-01-17
Add browser testing

beck:
## As-is

A bug was able to make it to release.

Applying these changes to tag v3.2.0, the bug would not have shipped because the tests would have failed with:

```
3) uuid() returns a 36 char string:
unknown error: Illegal invocation
running chrome
Error: An unknown server-side error occurred while processing the command.
    at execute("return uuid();") - index.js:312:3
```

## Changes

Add browser testing.  Looks like:
```
$ npm run test:browser

[chrome #0-0] Session ID: 369448ac883da88b9dbed611e56935f0
[chrome #0-0] Spec: /Users/doug/Projects/node-uuid/test/specs/index-spec.js
[chrome #0-0] Running: chrome
[chrome #0-0]
[chrome #0-0] uuid()
[chrome #0-0]   ✓ returns a 36 char string
[chrome #0-0]
[chrome #0-0]
[chrome #0-0] 1 passing (2s)
[chrome #0-0]
```

If unfamiliar with end-to-end testing, the scaffolding can be painful because three processes are needed:

1. http server, serving up the code under test
1. local browser driver (aka selenium) or a proxy server to drive cloud browsers
1. a test runner to orchestrate and feed in specs

I've left all source untouched. I've tried my best to keep the boilderplate minimal.

**Why webdriver.io?** 

In the past I've tried karma, grunt & gulp plugins, and nightwatch.js. They all had more scaffolding.

How `wdio` is configured, it is easy bake everything into one call.

Misc thoughts:
* webdriver.io API is synchronous, making it significantly easier to write clean tests.
* Mocha formatted tests seem to "just work" out of the box (not so with nightwatch)
* Custom config is easy to set up (includes a wizard) or extend/modify other config
* Nightwatch made it hard/impossible to skip tests

## What's next?

* create the account `node-uuid` on saucelabs

* update the project setting in travis (found under "more options" > "settings") and env vars:

  * `SAUCE_ACCESS_KEY`
  * `SAUCE_USERNAME`

* update `.travis.yml` to tie-in the tests

* most important step, add a badge to the readme. Everyone loves badges:

  ```
  [![Test Status](https://saucelabs.com/buildstatus/node-uuid)][sauce]
  [sauce]: https://saucelabs.com/u/node-uuid
  ```

## Caveats

The browser tests only need to run under one node environment, I see travis is configured for several.

These new dependencies will not play nice with `0.12`. ATM TravisCI [is backlogged](https://twitter.com/traviscistatus/status/953423338591854592) but I do have a fix waiting: https://github.com/beck/node-uuid/commit/55e3b3fbcfd2f6a7aa45b6dbb2fc16c588b6ded8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
+/wdio.*.js
 *.sw*
 .idea/**
 .DS_Store
 node_modules
+test/specs/*.bundle.js
+user.json
--- a/.npmignore
+++ b/.npmignore
@@ -3,3 +3,5 @@ test/
 .gitignore
 .travis.yml
 .npmignore
+user.json
+wdio.*.js
--- a/README.md
+++ b/README.md
@@ -286,8 +286,44 @@ Type `uuid --help` for usage details
 ## Testing
 
 ```shell
+# Run the unit tests and browser tests:
 npm test
+
+# Just the unit tests
+npm run test:unit
+
+# Just the browser tests
+npm run test:browser
 ```
+
+### Run the tests using cloud browsers
+
+Requires a [Sauce Labs][] username and access key.
+
+```
+echo '{"user": "<username>", "key": "<access key>" }' > user.json
+npm run test:sauce
+```
+
+[Sauce Labs]: https://saucelabs.com/
+
+### Run the tests using custom configuration
+
+Browser tests scaffolded upon [webdriver.io][].
+
+Create a `wdio.conf.js`, using the sample:
+```shell
+cp test/wdio.sample.js wdio.conf.js
+```
+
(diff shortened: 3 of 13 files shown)


Andarist:
Wouldnt a simpler setup with local puppeteer be "good enough"? I understand it wouldn't catch IE's `msCrypto`, but it also could be mocked in a test to reflect IE's behaviour.

ctavan:
We have added extensive browser tests in #348 that should give us some safety for the future. Feel free to build on that work if you think that anything is still missing.

beck:
Nice! 🎉

PR #270, opened 2018-04-05
fix assignment to readonly property to allow running in strict mode

### LINKED ISSUE #268: Can't run with --use_strict due to assignment of read-only function.name property in v35.js
https://github.com/kelektiv/node-uuid/blob/17d443f7d8cfb65a7d85803218df27a6b4ae8397/lib/v35.js#L46

The above line is assigning to a read-only variable, which is a violation when running in strict mode (via Node's `--use_strict` argument).

Outside of strict mode, this doesn't actually result in the function name being changed anyway, as assignments to `Function.name` are ignored.

There are [other ways to change the name of a function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name) that hopefully would work. 

If I comment that line out, everything works fine, with no other errors.

G-Rath:
Replaced assignment to readonly Function.name with Object.defineProperty, as you can't assign to readonly properties in strict mode.

Fixes #268

### DIFF
--- a/lib/v35.js
+++ b/lib/v35.js
@@ -43,7 +43,7 @@ module.exports = function(name, version, hashfunc) {
     return buf || bytesToUuid(bytes);
   };
 
-  generateUUID.name = name;
+  Object.defineProperty(generateUUID, 'name', {value: name});
 
   // Pre-defined namespaces, per Appendix C
   generateUUID.DNS = '6ba7b810-9dad-11d1-80b4-00c04fd430c8';
--- a/lib/v35.js
+++ b/lib/v35.js
@@ -43,7 +43,11 @@ module.exports = function(name, version, hashfunc) {
     return buf || bytesToUuid(bytes);
   };
 
-  Object.defineProperty(generateUUID, 'name', {value: name});
+
+  // only attempt to set the name if's configurable (which it should be on node > 0.12)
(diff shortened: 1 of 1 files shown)


G-Rath:
@broofa for some reason the `readme.md` was changed for me - It seems the uuids were regenerated?

Since I don't know how that file works, I've not committed the changes. Let me know if I should.

G-Rath:
@broofa it seems this solution won't work with [node 0.12.x](https://stackoverflow.com/questions/31977822/can-function-names-be-configured-in-node-js), as it's a non-standard, pre-ES2015 implementation.

I've fixed this by adding a conditional check to make sure the `name` property is actually configurable before trying to configure it.

Given that prior to this the assignment to `generateUUID.name` wasn't apparently doing anything anyway, I don't expect this to be a breaking change.

Let me know what you think, and if there are any adjustments you'd like me to make :)

BenjaminVanRyseghem:
I published https://www.npmjs.com/package/uuid-pull-270 on NPM based on @G-Rath branch in case someone need it as well 😄

johngeorgewright:
Hello @broofa, any news on when this will get published? Struggling to get node-uuid working without this patch.

PR #274, opened 2018-04-18
Get correct version of IE11 crypto

robwierzbowski:
Users on Windows 7 and IE11 sometimes trigger a bind error because the crypto object exists but getRandomValues is undefined. Using https://github.com/openpgpjs/openpgpjs/pull/207 as a reference, I've added a more robust check for the necessary crypto properties in IE 11.

We deployed our branch into production yesterday. I'll report back in a week to let everyone know if this fixed the issue or not, and if so we can potentially merge?

### DIFF
--- a/lib/rng-browser.js
+++ b/lib/rng-browser.js
@@ -3,9 +3,11 @@
 // and inconsistent support for the `crypto` API.  We do the best we can via
 // feature-detection
 
-// getRandomValues needs to be invoked in a context where "this" is a Crypto implementation.
-var getRandomValues = (typeof(crypto) != 'undefined' && crypto.getRandomValues.bind(crypto)) ||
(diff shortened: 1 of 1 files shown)


broofa:
Side note: Can't wait for the day [optional chaining](https://github.com/tc39/proposal-optional-chaining) is a thing.  :-p

robwierzbowski:
Commit message updated.

Re: optional chaining, for sure. Although I wish there was an alternate syntax, I'm starting to get PTSD flashbacks from working in CoffeeScript 😅

robwierzbowski:
Oh, btw we have not had any bind errors from node-uuid on IE 11 since we deployed this fork. I believe it works.

robwierzbowski:
@broofa Can you note when you release a new version on this message so we can pull the official package back into our dependencies?

PR #288, opened 2018-08-11
fix: add TypeScript defaults support

shellscape:
This PR adds a simple `module.exports.default` line to each of the version files.  When the following `tsconfig.json` values are used:

```json
    "module": "commonjs",
    "target": "es6"
```

And `esModuleInterop` is `false`, TypeScript will compile the uuid-consuming file with the assumption that `module.exports.defaults` exists. One might argue that this change is not needed, but `esModuleInterop` can really screw things up when using the compilation result in certain bundlers. Several other widely used CJS modules have also added this tiny bit of compatibility with no ill-effect.

### DIFF
--- a/v1.js
+++ b/v1.js
@@ -107,3 +107,4 @@ function v1(options, buf, offset) {
 }
 
 module.exports = v1;
+module.exports.default = v1;
--- a/v3.js
+++ b/v3.js
@@ -2,3 +2,4 @@ var v35 = require('./lib/v35.js');
 var md5 = require('./lib/md5');
 
 module.exports = v35('v3', 0x30, md5);
+module.exports.default = module.exports;
--- a/v4.js
+++ b/v4.js
@@ -27,3 +27,4 @@ function v4(options, buf, offset) {
 }
 
 module.exports = v4;
+module.exports.default = v4;
--- a/v5.js
+++ b/v5.js
@@ -1,3 +1,4 @@
 var v35 = require('./lib/v35.js');
 var sha1 = require('./lib/sha1');
 module.exports = v35('v5', 0x50, sha1);
+module.exports.default = module.exports;


broofa:
If the solution is to expect everyone else to add a line of code, it's not a solution.

I.e. My policy is to keep this module free of app/tool/framework-specific code.  If/when https://nodejs.org/api/esm.html gains traction, I'll port this code over to ES6 module style.  Until then, you'll need to find different workaround.

shellscape:
@broofa _please please_ reconsider. Your module is a staple dependency for so many people and this is such a small addition that adds major benefit. Adding a `default` export is commonplace these days. ESM has gained traction, and it's available behind a flag just as async/await was. I don't think anyone would argue that ESM has not infiltrated the mainstream. This being closed is such a massive bummer.

johnnyreilly:
Hey @broofa,

TypeScript user here!  Thanks for your mighty fine library.

> My policy is to keep this module free of app/tool/framework-specific code

I completely appreciate that you have a policy and why. Changes like this shouldn't be something that you arbitrarily add. That said, I'd say there's a large audience of people who'd be super grateful if you were willing to take this PR.

As @shellscape says:

> Several other widely used CJS modules have also added this tiny bit of compatibility with no ill-effect.

Would it help to convince you if he was to list some examples just so know this isn't a flash in the pan amend? Rather something that's being generally adopted and hence more likely to make you reconsider?

Obviously it's your joint and you get to decide; Your house, your rules and all that. But I wanted to check if you're open at all to this option. 

Thanks for your work!

broofa:
> Your module is a staple dependency for so many people

Hence, why it's better to not contaminate it with tooling-specific clutter.

While I appreciate your concern, I've been down this path in earlier versions of this module.  One of the best decisions @defunctzombie made when he forked this off my original repo was to kill [all the copy-pasta code](https://github.com/kelektiv/node-uuid/commit/978212cbcece56934a1a9f65c024d41f9c9dec39#diff-bef26ab2ce6b354722523f65bc301d59L252) I had for supporting the various module loading schemes (AMD, RequireJS, CommonJS, et al).  By standardizing on CommonJS and putting the responsibility for loading onto the packager, it made it possible to clean up the code and dependencies.

> Several other widely used CJS modules

Sure... but where do you draw the line? JS dependencies are notoriously [convoluted](http://npm.broofa.com/?q=npm).  Are you going to persuade the whole ecosystem to make this change?  Surely there's a better way.

Be aware that TypeScript is just the latest in a long list of tooling that claims to "solve" some core JS issue (often module loading) in an ad-hoc manner.  Component.js, Webkit, Babel, Angular, Browserify, Coffeescript, ES6/7/8, Grunt, Gulp, Yeoman, Brunch, JSPM, Parcel... etc., all play in this space. And because this is such a widely used module, it's used in pretty much every possible combination of ways, which simply can't be tested effectively.  While you claim your change is harmless, I can pretty much guarantee you it'll have unintended consequences _somewhere_.  And guess who ends up having to deal with that issue when it gets filed?

My suggestion would be to create your own module that wraps this one and exposes the `default` property as a temporary(?) fix.  Then get behind a proposal for standardizing the `default` property that is adopted across all of the packager tools out there (akin to the `[browser](https://github.com/defunctzombie/package-browser-field-spec)` package.json property that @defunctzombie put together).  If that gains traction, then I'll reconsider.

PR #290, opened 2018-09-02
Added OID namespace for v3/v5 UUIDs

rhowell:
Added the ability to generate v3 & v5 UUIDs using the OID namespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ const uuidv3 = require('uuid/v3');
 // ... using predefined DNS namespace (for domain names)
 uuidv3('hello.example.com', uuidv3.DNS); // ⇨ '9125a8dc-52ee-365b-a5aa-81b0b3681cf6'
 
+// ... using predefined OID namespace (for domain names)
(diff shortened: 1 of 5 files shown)


broofa:
Hmm, interesting.  What's in those CSV files that requires an OID UUID, if you don't mind my asking?  Can you elaborate a bit on what you're doing?

I've never had a use case for OID UUIDs, and have only skimmed the OID docs.  Truth be told, they're pretty esoteric and I don't completely understand what the use cases are.

rhowell:
They have ISO country and administrative areas (states/provinces/etc...).

broofa:
What's the connection between ISO country/admin area and OID UUIDs ?   E.g. what OID "arc" do these fall under?  I know it's none of my business and not germane to whether this module should include an OID constant.  I'm just curious.

While I've drawn a clear line with this module about strictly adhering to RFC4122, the OID namespace is only mentioned once, [in Appendix C](https://tools.ietf.org/html/rfc4122#appendix-C), as a "potentially interesting name space".  And in researching OIDs I'm [still] of the impression that they're  rather esoteric and likely to remain so.  Thus, I think it best to leave the OID constant out and suggest that people like yourself simply define it within your own code.

I remain open to arguments to the contrary however.

PR #294, opened 2018-10-04
Throwing error when insecure rng is used

### LINKED ISSUE #173: Insecure random values should be opt-in

uri:
- New option for v1 and v4: allowInsecureRng
  - Require to be `true` if `rng.insecure == true`

Closes #173

----

Here's my attempt at closing #173, please let me know if a different approach is preferred.

### DIFF
--- a/lib/rng-browser.js
+++ b/lib/rng-browser.js
@@ -17,18 +17,7 @@ if (getRandomValues) {
     return rnds8;
   };
 } else {
-  // Math.random()-based (RNG)
-  //
-  // If all else fails, use Math.random().  It's fast, but is of unspecified
-  // quality.
-  var rnds = new Array(16);
-
-  module.exports = function mathRNG() {
-    for (var i = 0, r; i < 16; i++) {
-      if ((i & 0x03) === 0) r = Math.random() * 0x100000000;
-      rnds[i] = r >>> ((i & 0x03) << 3) & 0xff;
-    }
-
-    return rnds;
+  module.exports = function noSecureRNG() {
+    throw Error('uuid: No secure RNG available. See https://github.com/kelektiv/node-uuid/wiki#no-good-rng-error for more info.');
   };
 }
--- a/test/test.js
+++ b/test/test.js
@@ -60,16 +60,9 @@ test('nodeRNG', function() {
   }
 });
 
-test('mathRNG', function() {
+test('noSecureRNG', function() {
   var rng = require('../lib/rng-browser');
-  assert.equal(rng.name, 'mathRNG');
-
-  var bytes = rng();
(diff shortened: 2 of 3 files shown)


broofa:
At this point I'm of the opinion this module shouldn't have an insecure option.  Thus ...

1.  Get rid of the `insecure` and `insecureMessage` stuff
2. Remove the checks in `v1` and `v4` (they're RNG agnostic and should stay that way)
3. Change `mathRNG()` to `noSecureRNG()`, thusly: 
```
function noSecureRNG() {
    throw Error('uuid: No secure RNG available. See <github link> for more info.');
}
```
4. Add an `rng` option to `v1`, similar to `v4`, so users can pass a custom RNG if desired as a workaround to the error being thrown. 

The link in step 3 should refer either to docs in the README or a separate wiki page in this project that describe:
* Why the error occurs ("You're running on a JS platform that doesn't provide a secure RNG")
* How to work around it ("Pass an `options.rng` param into either v1() or v4()`")
* An example of how to do that.

That last point is a little problematic. Ideally we could recommend a cryptographically secure JS module upon which to build a CSPRNG function the reader could pass in, but that's easier said than done. Such modules tend to be either obscenely code-heavy or of questionable quality, and there's no obvious "right" choice.  If we make a recommendation, we (meaning "I") end up dealing with any issues that arise, which I'm not interested in taking on.  So... I think a simple example that just shows the basic mechism for passing in an rng function is sufficient.  It could use an all-zeroes function as an example (`function nilRNG() {return new Array(16).fill(0)}`).  What it should _not_ do is provide an example that uses Math.random(), as that's just perpetuating the problem.

broofa:
BTW, thanks for working on this! :-)

uri:
Thanks for the feedback. I'll make those requested changes.

uri:
@broofa I made those changes and modified https://github.com/kelektiv/node-uuid/wiki#no-secure-rng-error

lime:
Hi! I was wondering if I can be of any help in getting this across the finish line?

It seems like the only things missing are an amended commit message and a changed wiki link. I forked this branch and made those changes here: https://github.com/lime/node-uuid/tree/error-on-insecure-rng

If you want, I could open a separate pull request from it. You're also welcome to pluck those commits in directly. I just wanted to check first whether it would be welcomed, and if there's anything else that still needs work.

uri:
@broofa I made the change to the commit message is there anything else?

ctavan:
Closed through #354

PR #295, opened 2018-10-05
Add missing namespaces to v35.js

CaptainVascular:
Add missing predefined namespaces for OID and X500 from the RFC 4122 spec to v35.js
See https://tools.ietf.org/html/rfc4122#appendix-C


broofa:
See #205

PR #304, opened 2019-01-31
Export version 3 & 5 uuids

uakb4r:
The implementation is there but export is missing


broofa:
This is intentional.  Per the deprecation warning in the README, users are discouraged using the API on the `uuid` object.  v1 and v4 are only available there for backward-compatibility reasons.

PR #306, opened 2019-03-07
June '19 update

broofa:
Fixes vulnerability warnings.
Keeps README from changing every time it's generated
Remove node v4 from travis (`mocha` no longer supports v4)


.npmrc
```diff
@@ -0,0 +1,2 @@
+tag-version-prefix="v"
+save-prefix="~"
```
defunctzombie: would suggest exact versions rather than any floating version - but your call

broofa: 1. `package.json` already specifies the exact versions, and is what `npm install` prefers (I believe).
2. I actually see exact versions in package.json as an anti-pattern, as it makes it redundant with  `package-lock.json`, while nullifying any value to be had from the `npm outdated` and `npm upgrade` commands.

broofa:
@defunctzombie Updated to address recent github audit warnings.  Approve, please.

package.json
```diff
@@ -17,16 +17,16 @@
...
   },
   "devDependencies": {
-    "@commitlint/cli": "7.0.0",
-    "@commitlint/config-conventional": "7.0.1",
-    "eslint": "4.19.1",
-    "husky": "0.14.3",
-    "mocha": "5.2.0",
-    "runmd": "1.0.1",
-    "standard-version": "4.4.0"
+    "@commitlint/cli": "7",
```
defunctzombie: Personally not a fan of this style of dependency tracking (for same reasons as mentioned earlier in this PR).

PR #307, opened 2019-03-27
Add ES6 import example to readme

killthekitten:
This is to remove the confusion that is [certainly present](https://github.com/kelektiv/node-uuid/issues/245) at the moment.


broofa:
Thanks for contributing.  A couple issues, however ...

1. Edit README_js.md for docs, not README.md.
2. Your code sample is incorrect.  Correct sample [here](https://github.com/kelektiv/node-uuid/issues/245#issuecomment-355418687).

Issue #245, opened 2017-12-21
ES6 import support?


Samstiles:
Hey y'all,

Thanks for your work on this lib. Any plans for ES6 import support in the near future? If it's something actively being worked on I'm gonna wait, otherwise I may look into possibly PRing?

EG

`import { uuidv1 } from 'node-uuid'`

ZachMayer35:
not quite what your example asks for, but you can already do this:

`import {v4 as uuidv4} from 'uuid';`

or just

`import {v4} from 'uuid';`

if you're not picky about the accessor.

broofa:
As [noted in the README](https://github.com/kelektiv/node-uuid#deprecated-api),  importing `uuid` is deprecated.  Instead, import `uuid/<version>`.   E.g. to import the v1 generator using ES6 notation:

`import uuidv1 from 'uuid/v1';`

(Note: I haven't tested this, but give it a try and report back here with the results.  I'll reopen if this is a legit issue.)

Oliboy50:
@broofa `import uuid from 'uuid/v4';` syntax does not work, at least in a Typescript v3 project running on Node v10.9.0

on the other hand, `import {v4 as uuid} from 'uuid';` works as expected

broofa:
Apologies for the late response on this.  I [still] consider this issue closed, but in light of the "reactions" the above comment has garnered, let me state my position on this...

> at least in a Typescript v3 project 

... which makes this a Typescript (or TS configuration) issue.  This module conforms with the documented CommonJS and ES6 behaviors.  You can use ES6 syntax either directly in the browser or via webpack w/out issue, so whatever the issue is here, it's specific to TS.

Oliboy50:
@broofa actually you're right, in my case this was only due to how `Typescript` is handling ES6 imports

so, just like many other JS libs like this one, it works this way using TS:

```ts
import * as uuid from "uuid/v4";

console.log(uuid());
```

here is a codesandbox to play with: https://codesandbox.io/s/5372omm68n

jonathanstiansen:
@Sampath-Lokuge that doesn't work in Typescript.

Sampath-Lokuge:
@jonathanstiansen What was the error? I have used above on Ionic / Typescript app and no issues.

MartinX3:
Typescript 3.2.4 in angular 7.2.7

Creating an object from this class
```
import { UUID } from 'uuid/v4';

export class entity {
  id: string;
  name: string;

  constructor() {
[...]

PR #317, opened 2019-06-25
Experimental rewrite: Es6 modules, BigInt, (un)parsing and validation

broofa:
#245 and as well as some of the work we're doing over at https://github.com/bcoe/proposal-standard-module-uuid have had me playing around with what this module might look like if it was rewritten using modern ES features (ES6 modules, BigInt) and with a UUID class for parsing/unparsing and validation.

This is the work-in-progress result.  (Posting this mostly because my laptop's logic board decided to crap out and enough work went into it I'd rather not risk losing it while dealing with Apple repair process.)

Feel free to comment, with the caveat that I don't have a concrete plan (yet) for merging to master here.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,9 +7,12 @@
     "node": true,
     "mocha": true
   },
+  "globals": {
+    "msCrypto": true
+  },
   "parserOptions": {
     "sourceType": "module",
-    "ecmaFeatures": {"ecmaVersion": 6}
+    "ecmaVersion": 2017
   },
   "extends": ["eslint:recommended"],
   "rules": {
@@ -22,6 +25,7 @@
     "comma-spacing": ["warn", {"after": true}],
     "dot-notation": "warn",
     "eqeqeq": ["warn", "smart"],
+    "func-call-spacing": ["warn", "never"],
     "indent": ["warn", 2, {
       "SwitchCase": 1,
       "FunctionDeclaration": {"parameters": 1},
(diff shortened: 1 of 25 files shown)


DarylThayil:
how far off is this from landing :D

broofa:
@DarylThayil: As noted, there's no set plan for landing this.  This is just a my take on what the codebase for this [9-year old module] would look like if it was updated to leverage new and coming language features.  I don't know that there's a lot of demand for this, either.  While it does address a few of the longer-running issues we've had with uuid (ES6 modules, parsing, validation support), none of those are critical.

Probably the biggest issue with this branch is the use of `BigInt`.  It's [not supported on Safari or Edge](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#Browser_compatibility), nor is there a suitable polyfill for it that I'm aware of.

43081j:
to be honest a standalone ESM rewrite would do wonders (without the bigint changes, as they should probably be delivered separately).

if there's no plan to at least do the ESM side of things im tempted to fork for now.

in its current state, uuid can't be consumed by es modules AFAIK and can't be used in TS projects targeting ESM.

i'd personally go as far as saying this needs to be split into a node-only build and a browser-build, possibly in separate repos.

broofa:
Closing.  The UUID class is interesting, maybe as a separate module in the uuidjs org or something, but not relevant to current development work.

ctavan:
@43081j @DarylThayil @dstaley:

Sorry for taking such a long time but we have finally released `uuid@7.0.0-beta.0` which features native ESModules-Support for the browser:

```
npm install uuid@7.0.0-beta.0
```

The [master branch README](https://github.com/uuidjs/uuid/blob/master/README.md) already reflects the API of v7.0.0.

It would be awesome to get feedback from you if this release works well for you.

PR #319, opened 2019-07-23
CLI: encoded/binary input (v3/5)

tschiemer:
Hi! I see the buffer/array support was added.

I would (again, see #216) propose to add the option to pass encoded (in particular binary) data to the CLI.

Previously you considered the CLI a convenience for simple use cases and binary names would not fall into that category.

When working with unique hardware ids or serial numbers, arbitrary byte sequences occur that would have be to be transformed otherwise beforehand.

Although for such cases, custom scripts may exist, but this feature might still be a nice to have out-of-the-box convenience - in particular when having to manually test or verify other processing steps.

Merely proposing ;)


ctavan:
@tschiemer sorry for the vary late response to this proposal.

Is this still a relevant use case for you? If yes, would you be willing to update the pull request?

We have been reworking this library quite a bit, so this PR needs a rebase…

@broofa would you be open to accept the cli change now as it only affects v3/v5 and seems to provide a reasonable use case for those in need of v3/v5?

bin/uuid
```diff
@@ -49,6 +59,14 @@ switch (version) {
     var uuidV5 = require('../v5');
 
     var name = args.shift();
+
+    // if two more arguments remain, see if an ecoding has been specified in place of the expected name and process accordingly
+    if (args.length == 2 && /^--encoding=([a-z0-9]+)$/.test(name)){
+      var newBuffer = (typeof Buffer.from === 'function' ? function(val,format){ return Buffer.from(val,format); } : function(val,format){ return new Buffer(val,format); });
```
ctavan: We have dropped support for Node.js v0.4 so please rely `Buffer.from`)

bin/uuid
```diff
@@ -30,6 +32,14 @@ switch (version) {
     var uuidV3 = require('../v3');
 
     var name = args.shift();
+
+    // if two more arguments remain, see if an ecoding has been specified in place of the expected name and process accordingly
+    if (args.length == 2 && /^--encoding=([a-z0-9]+)$/.test(name)){
```
ctavan: The regex is used multiple times. Consider defining it as a variable in the beginning.

bin/uuid
```diff
@@ -30,6 +32,14 @@ switch (version) {
...
+      var newBuffer = (typeof Buffer.from === 'function' ? function(val,format){ return Buffer.from(val,format); } : function(val,format){ return new Buffer(val,format); });
+      var format = name.match(/^--encoding=([a-z0-9]+)$/)[1];
```
ctavan: Why invent a new name and not simply call this variable `encoding`?

bin/uuid
```diff
@@ -30,6 +32,14 @@ switch (version) {
...
+      name = Array.from(newBuffer(args.shift(), format));
```
ctavan: For better readability maybe something like:

```js
var encoding = name.match(/^--encoding=([a-z0-9]+)$/)[1];
var encodedName = args.shift();
name = Array.from(Buffer.from(encodedName, encoding));
```

broofa:
I'm still not convinced this is needed.  It increases the surface of the CLI and README, for a feature that only one person has asked for.  Kinda seems like we should have a higher bar for adding features at this point.

(That said, I've submitted my fair share of PRs to other projects where I seem to be the only one who cares so... note my hypocrisy, I guess.)

ctavan:
@broofa I agree with having a rather high bar for feature additions since this module is so widely depended upon and we should only add stuff to it that is also widely being asked for. Also the CLI is definitely not the main purpose of this module.

@tschiemer if you really primarily need a commandline tool for your specific use case you could also create a separate module that uses this library under the hood. You will then have the benefit of having a solid implementation under the hood but can still design commandline tool to suit your needs.

If there are more folks who have @tschiemer's use case please chime in and we'll reconsider adding this feature.

PR #327, opened 2019-09-21
Specify all modules at main of the library

Seia-Soto:
I think the library needs to support multiple versions of UUID at one time without requiring more than 2 times. I am using multiple versions at one time, and now there are more lines in my code. It is not a big deal for me. However, I thought if I can do like below(code), it is really helpful.

```javascript
const uuid = require('uuid')

const v1Result = uuid.v1(); // to use v1
uuid.v5('main', v1Result); // use v5 right after v1
```

According to #154, the versions were separated into submodules and the structure of the library was changed. The only change I want is specifying all submodules at the main(index.js). From structure, I think there are no changes to use this library.

### DIFF
--- a/README_js.md
+++ b/README_js.md
@@ -89,6 +89,25 @@ const MY_NAMESPACE = '1b671a64-40d5-491e-99b0-da01ff1f3341';
 uuidv5('Hello, World!', MY_NAMESPACE); // RESULT
 ```
 
+To create multiple versions of UUID at one time:
+
+```javascript
+const uuid = require('uuid');
+
+uuid.v1(); // ⇨ '2c5ea4c0-4067-11e9-8bad-9b1deb4d3b7d'
+uuid.v4(); // ⇨ '1b9d6bcd-bbfd-4b2d-9b5d-ab8dfbbd4bed'
+
+const MY_NAMESPACE = '1b671a64-40d5-491e-99b0-da01ff1f3341';
+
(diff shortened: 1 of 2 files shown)


broofa:
80+% of [opensource] projects using this library use v4 and nothing else.  Adding dependencies to the top-level module increases the footprint of this module for those users by 6-8X.  Unless/until packagers are able to effectively "tree shake" the unused code out of bundles (which, to my knowledge, they do not currently do), I would suggest we not make this change.

@ctavan may have an opinion on this, based on his recent ESM work.

ctavan:
@broofa I believe once we land an ESM variant of this library it will be safe to export all versions from the top-level, at least my experiments with `webpack` indicated that: https://github.com/kelektiv/node-uuid/tree/a66f7935eb060eb963b6dc21ab90f015cbb078dc/examples/browser-webpack/dist

We could then also include all versions in the CommonJS-build assuming this build will mainly target Node.js users but I think that is a separate discussion.

Until we have a viable alternative for the Web platform I agree with you that we should not further in crease the footprint of the CommonJS-module variant of this module.

broofa:
Yeah, that's pretty much my understanding - that tree-shaking is feasible with ESM modules, but isn't practical for CommonJS codebases.

Seia-Soto:
I see why there aren't all versions at the main entry. I am sorry to do this work without considering other uses because I am not good at front-end technologies such as webpack. I can't talk pretty enough about this subject but now understood why. Thanks!

PR #335, opened 2019-10-14
Prepare for esm

ctavan:
The preparational commits from #331 extracted for better reviewability.

This should be a noop change compared to the currently released npm package.


ctavan:
@broofa can we merge these changes into master before continuing the efforts on the next branch?

PR #341, opened 2019-10-25
Modernize test tooling

ctavan:
Follow-up PR to #347


ctavan:
@broofa I'm not willing to maintain a consistent code formatting on my own that's why I propose to add prettier which, in my experience, does a pretty good job at maintaining consistent formatting.

Thanks to `eslint-plugin-prettier` it also plays rather well with `eslint`.

PR #347, opened 2019-12-03
Use GitHub actions

ctavan:


ctavan:
@broofa @defunctzombie I stopped trying to figure out how I could change the Travis.CI integration and instead simply moved over to GitHub actions.

They seem to execute even faster than Travis and I find them easier to maintain. I don't see any drawback.

Fine for you?

defunctzombie:
Late to the party but I am 👍 on GH actions.

PR #352, opened 2020-01-16
Prepare for canary release

ctavan:
This is a collection of a few more cleanup and tooling commits in preparation of a first 4.0.0-canary release.


ctavan:
As stated in the [commit message](https://github.com/uuidjs/uuid/pull/352/commits/1a644492cc9388216476d9a547bf101a280567ad) this is really mostly personal preference.

I find `yarn upgrade --interactive` a really helpful tool and I still couldn't find a true alternative with npm (i.e. upgrade dependencies in the lockfile while ALSO upgrading the package.json to reflect the latest version choices AND not having to manually specify the desired versions), the only thing I'm aware of is https://www.npmjs.com/package/npm-check-updates

Do you have strong feelings against yarn?

broofa:
My main concern with yarn is that it's yet another dependency in the toolchain that consumers / contributors need to be familiar with if they want to dive into the project.  It's not that big a deal since it's only surfacing in the README as part of the build instructions for the examples and testing.

I guess my preference would be to avoid it unless it's adding real, tangible value, but I'll leave that up to you.  I recognize that this is something I'm probably more sensitive to than most folks. :-)

ctavan:
@broofa after getting more familiar with `npm` and `ncu` I think I can also live with it. So I reverted the introduction of yarn!

Moreover I have changed and documented the `npm run release` toolchain slightly to respect the fact that we now have a build step and that the resulting files will be located in `./dist` and not in the top level of the module. I also make sure to only publish the dist files (+README/CHANGELOG/LICENSE) now using the `files` whitelist in `package.json` instead of the `.npmignore` blacklist.

I would now go ahead and introduce the breaking change of removing insecure RNGs (#173) and then put out a first `4.0.0-canary.0` under a different dist tag release. Fine for you?

Would like to put the functional changes into yet another PR though.

package.json
```diff
@@ -14,33 +14,39 @@
...
-    "http-server": "~0.12.0",
-    "husky": "~3.0.9",
-    "jest": "~24.9.0",
-    "lint-staged": "~9.4.3",
-    "npm-run-all": "~4.1.5",
-    "prettier": "~1.18.2",
-    "runmd": "1.2.1",
-    "selenium-webdriver": "~3.6.0",
-    "standard-version": "7.0.0"
+    "@babel/cli": "^7.8.3",
```
broofa: Change .npmrc to have `save-prefix="^"`

(For the record, I'd be okay removing version prefixes altogether so we have explicit versions throughout.)

ctavan: I switched to explicit versions.

README_js.md
```diff
@@ -297,3 +297,13 @@ Type `uuid --help` for usage details
 ```shell
 npm test
 ```
+
+## Releasing
```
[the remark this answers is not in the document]
ctavan: Following this logic, the documentation on how to run the unit tests also wouldn't belong into the readme.

I moved both into the new `CONTRIBUTING.md` file. I personally find it hard to impossible to keep the wiki in sync with the code and would therefore prefer to keep such documentation within the repo.

PR #353, opened 2020-01-20
Add bundlesize

ctavan:
This will let us prevent regressions in bundlesize.

I chose to monitor the bundlesize of webpack and rollup build because this is what ultimately matters to the user. If we break something in our library that would e.g. break tree shaking in one of these bundlers we would want to notice.

Closes #355


ctavan:
@broofa how do you feel about adding bundlesize to this repo (see the additional check in "show all checks")?

broofa:
If this is based on #352, can you change the base to `prepare-for-alpha-release` so it's easier to review, please?  (That makes sense, right?)

ctavan:
Uhm yeah, sorry. I rebased now, should be easy to review.

ctavan:
Sorry for the noise of the many force pushes. I was experiencing flaky browser tests here. It seems that _sometimes_ [this await](https://github.com/uuidjs/uuid/blob/380fe2844098f5db5438c8a497efc30e1f9c6252/test/browser/browser-test.js#L127) just never returns and the test just stalls there until it hits the jest timeout. I have no clue why, it seems to happen across almost all browsers (haven't seen it in Chrome so far). I'm a bit clueless 🤷‍♂ . Best bet so far is to just re-run the browser tests through the GitHub actions UI until they don't time out.

Anyways, the bundlesize part in this PR is ready for review now.

ctavan:
@broofa thanks for the suggestion. You are right, it is better to check the individual versions!

However I also want to ensure that we don't break the treeshaking capabilities which is why I don't want to specify the source esmodule files as entrypoints to rollup and webpack, but instead use separate files that explicitly do:

```js
import { v1 as uuidv1 } from 'uuid';
```

Fine for you?

(I picked your commit and added an additional change commit, would squash merge everything in here in the end)

PR #354, opened 2020-01-20
Remove insecure rng

### LINKED ISSUE #173: Insecure random values should be opt-in

ctavan:
Closes #294 and fixes #173


ctavan:
BTW as expected this now fails in IE 10 where no crypto.getRandomValues() is available:

Screenshot 2020-01-20 23 35 52

ctavan:
@broofa do you think that, given the now-wide-spread support for CSRNGs in all platforms, we still need the wiki page over at https://github.com/uuidjs/uuid/wiki#no-secure-rng-error ?

broofa:
Go ahead and kill that wiki page.

ctavan:
@broofa did you close this PR on purpose?

broofa:
I did not.  Sorry about that!

README_js.md
```diff
@@ -123,8 +123,66 @@ npm run build
...
+
+```javascript --run v4
+const buffer = new Array();
+uuid.v4(null, buffer, 0); // RESULT
+uuid.v4(null, buffer, 16); // RESULT
+```
+
 ### Version 1 (Timestamp + Node)
 
+⚠️⚠️⚠️ **Please make sure to check whether you really need the timestamp properties of Version 1
```
broofa: Would be good to have a link of some sort that provides information to help with readers' v1 vs v4 decision making.

ctavan: I've added a link to the standard lib proposal. I think that covers the reasoning.

PR #355, opened 2020-01-22
improvement: check individual version bundle sizes

broofa:
Check separate version bundle sizes rather than combined size.


ctavan:
Was merged in #355

Issue #345, opened 2019-11-17
Allow using the generator function on a HTML page directly


prantlf:
The UUID generator can be built to a browser application using browserify, webpack, rollup or similar other bundler. However, if the application is built from pure JavaScript modules or [AMD] modules, this package cannot be used without an additional bundling work. (Scripts published from the NPM package cannot be loaded by the `script` element on a web page.)

How about including [UMD] modules in the NPM package, so that browser applications could load them directly, without each of them building such module themselves internally?

For example, by importing the UUID generator function globally:

```html
<script src="https://unpkg.com/uuid@3.4.0/browser/v1.min.js"></script>
<script>
  var id = uuidv1();
  console.log(id);
</script>
```

Another example, requiring the UUID generator as an [AMD] module (using [RequireJS], for example):

```html
<script src="https://unpkg.com/requirejs@2.3.6/require.js"></script>
<script>
  require.config({
    paths: {uuidv4: 'https://unpkg.com/uuid@3.4.0/browser/v4.min'}
  });
  require(['uuidv4'], function(uuid) {
    var id = uuid();
    console.log(id);
  });
</script>
```

[AMD]: https://github.com/amdjs/amdjs-api/blob/master/AMD.md#amd
[RequireJS]: https://requirejs.org/
[UMD]: https://github.com/umdjs/umd#umd-universal-module-definition

ctavan:
@prantlf I'm considering to add a UMD build when releasing the v4.0.0 version of this module so that this library could be loaded directly through the various CDNs that just mirror npm package content.

Until then could you check if https://github.com/kelektiv/node-uuid/issues/302#issuecomment-524827935 works for you?

ctavan:
This has been merged to the next branch with #357 and will be released with the upcoming major version release.

ctavan:
We have just released [`uuid@7.0.0`](https://github.com/uuidjs/uuid/blob/v7.0.0/CHANGELOG.md#700-2020-02-24) which now features direct UMD builds again: https://github.com/uuidjs/uuid#umd-builds

Please feel free to open a new issue if you encounter any issues with the new version.

Issue #333, opened 2019-10-13
feature request: CDN (again)


cekvenich:
The unstable CDN was removed.
If we can just have the compiled es5 .js file in the repo, then any CDN would work after an npm publish or github release. CDN's use npm or github release as source.
Then we users of the lib can pick a CDN we like.

So: please include a compiled es5 .js file in the repo; and release either npm publish or github publish please.

ctavan:
@cekvenich have you seen https://github.com/kelektiv/node-uuid/issues/302#issuecomment-524827935? Does this not work for you?

cekvenich:
it does, thx

Issue #302, opened 2019-01-19
wzrd.in service does not work


mostafa:
Build on the wzrd.in website seems not to work, because the whole server timed out! And building the package using browserify.js also does not produce a firm result and does not work.

jimaek:
jsDelivr also minifies the files along many other features https://cdn.jsdelivr.net/npm/node-uuid@1.4.8/uuid.min.js

Annis-Monadjem:
Why, uuid/v4 is not working ? What can be done ?

trs-king:
@Annis-Monadjem Yes, you can use uuidv4. 

Simply do `uuid.v4();`

mostafa:
> 
> 
> @mostafa: I'm interested in uuid version 4. Will your suggestion above: https://cdn.jsdelivr.net/npm/node-uuid@1.4.8/uuid.min.js produce uuidv4?
> 
> @jimaek: I'm interested in uuid version 4. Will your suggestion above: https://cdn.jsdelivr.net/npm/node-uuid@1.4.8/uuid.min.js produce uuidv4?
> 
> At present wzrd.in is crushing uuid/v4 which i'v been using with browserify.

It will, since I used and tested it. You can simply load it via an HTML boilerplate and try the console to inspect the uuid object.

Annis-Monadjem:
What i finally did was:
1)  $>yarn add uuid
2) In .js file:
import uuidv4 from 'uuid/v4'

Thanks @mostafa @trs-king

mostafa:
@Annis-Monadjem This jsdelivr link is for cases where you don't have any packer/bundler and are using browser javascript.

onlyanegg:
I'm also having problems with wzrd.in. I'm getting redirected to https, and then I'm getting a 502. I tried it yesterday and got the same result.

UkrainianCitizen:
So is this going somewhere? This library was referenced in JS course and all people including myself can't use browser-ready versions because of bad gateway error on wzrd.in.

ctavan:
@UkrainianCitizen as mentioned in #176 this is not something the maintainers of this module can change since apparently the service wzrd.in is generally down for over a year.

Please get in touch with the authors of your JS course in order to update the course material to use services that are still up and running.

If you need something that works at this point in time you can try including the following instead in your JS course HTML page:

```html
<script src="https://bundle.run/uuid@latest?name=uuid"></script>
```

[...]

PR #357, opened 2020-01-27
feat: add UMD build to npm package

ctavan:
A request for this has popped up a number of times in the issues.

I'm still not sure whether I want to add this. It adds another 28kb to the npm package and I'm not sure whether it's worth supporting such a legacy use case.

Plus there are options like https://bundle.run/uuid@latest?name=uuid

What do you think @broofa ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,5 +317,31 @@ $ uuid v1
 
 Type `uuid --help` for usage details
 
+## UMD Build
+
+If you want to load a minified UMD build directly in the browser you can use one of the popular npm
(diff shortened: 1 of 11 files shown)


broofa:
No strong opinions on this.  As long as it doesn't add to the maintenance headache I don't care much one way or the other.

ctavan:
Thought about it again and I think it's an acceptable overhead for a feature that has been requested at least three times last year:

- #302 
- #333 
- #345

PR #362, opened 2020-02-07
Replace bundlesize with bundlewatch

ctavan:
Unfortunately bundlewatch also doesn't seem to be super actively maintained however at least the server-side service seems to be working at the moment so we git diffs w.r.t. master directly in the Github check.

Since I can't find any drawbacks w.r.t. bundlesize I'd give it a try.


ctavan:
I'm watching http://npm.broofa.com/?q=@ampproject/filesize as well which has an even smaller dependency graph, however it doesn't support reporting differences w.r.t. master out of the box yet.

ctavan:
BTW apparently during the last few hours there was a problem on GitHubs side with verifying signed commits which has now been resolved, however commits need to be rewritten in order for GitHub to pick up the changes. This is why I had to rebase once more…

PR #363, opened 2020-02-13
docs: prefer esmodule import syntax in README

ctavan:
Especially for use in the browser modern ESModule import syntax is
preferable because it allows for proper treeshaking by bundlers like
Webpack/rollup.


ctavan:
@broofa I found a rather naive workaround at https://github.com/broofa/runmd/pull/2 to be able to continue to use runmd for this project.

PR #374, opened 2020-02-21
chore: rearrange/edit README

broofa:
Did and edit pass on the README.  Specifically:

* Updated the features list at the top
* Moved Quick Start to the top, since that is what most people are going to be interested in.
* Move release notes to the bottom.  Kept a "Upgrading to uuid@3?" line at the top directing people to them, however.
* Renamed all "v3.x"-style uuid version references to "uuid@3" to minimize confusion between RFC UUID "version" and uuid module version.  I.e Don't use "version" to refer to anything but RFC UUID version.
* Unicode warning symbols as html entity char codes (prefer not having actual unicode chars in source).  Also, the "⚠️ ⚠️ ⚠️ ... WARNING HERE ...⚠️ ⚠️ ⚠️" felt a bit excessive. 😄

### DIFF
--- a/README_js.md
+++ b/README_js.md
@@ -24,7 +24,7 @@ For the creation of [RFC4122](http://www.ietf.org/rfc/rfc4122.txt) UUIDs
   - CommonJS, [ECMAScript Modules](#ecmascript-modules) and UMD builds
   - Node 8, 10, 12
   - Chrome, Safari, Firefox, Edge, IE 11 browsers
-  - Webpack, Rollup packagers
+  - Webpack and rollup.js module bundlers
 - **Secure** - Cryptographically-strong random values
 - **Small** - Zero-dependency, small footprint, plays nice with "tree shaking" packagers
 - **CLI** - Includes the [`uuid` command line](#command-line) utility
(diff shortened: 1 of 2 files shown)


README_js.md
```diff
@@ -17,157 +17,77 @@ require('crypto').randomBytes = function() {
...
-Simple, fast generation of [RFC4122](http://www.ietf.org/rfc/rfc4122.txt) UUIDs.
+For the creation of [RFC4122](http://www.ietf.org/rfc/rfc4122.txt) UUIDs
 
-Features:
+- **Complete** - Support for RFC4122 version 1, 3, 4, and 5 UUIDs
+- **Cross-platform** - Support for ...
+  - CommonJS, [ECMAScript Modules](#ecmascript-modules) and UMD builds
+  - Node 8, 10, 12
+  - Chrome, Safari, Firefox, Edge, IE 11 browsers
+  - Webpack, Rollup packagers
```
ctavan: Both, webpack and rollup call themselves "module bundlers" (not packagers). Also we may want to use the official names (Webpack, rollup.js)

README_js.md
```diff
@@ -17,157 +17,77 @@ require('crypto').randomBytes = function() {
...
 ```
 
-Then generate a random UUID (v4 algorithm), which is almost always what you want ...
+Once installed, decide which type of UUID you need. RFC4122 provides for four
+versions, all of which are supported here. In order of popularity, they are:
+
+- Version 4 (random) - Created from cryptographically-strong random values
+- Version 1 (timestamp) - Created from the system clock (plus random values)
+- Version 5 (namespace, SHA-1) - Created from user-supplied name and namespace strings
+- Version 3 (namespace, MD5) - Like version 5, above, but with a poorer hash algorithm
```
ctavan: Link to the API docs below?

ctavan:
Also, I just merged #373 and a rebase is now necessary, sorry!

README_js.md
broofa: I don't think this is needed.  This is in the context of the "Quick Start", which provides examples immediately below this.  Pointing people at the detailed (more confusing) API docs doesn't feel helpful here.

broofa:
Failing test is unrelated.  Merging.

Issue #378, opened 2020-02-24
Version 7.0.0 not compatible with Azure Functions/AWS Lambda/Google Cloud Functions


ctavan:
References:
- https://github.com/uuidjs/uuid/issues/375#issuecomment-590428623
- https://github.com/uuidjs/uuid/issues/375#issuecomment-590576023

evdama:
It's not just Azure. AWS lambda as well as google cloud functions have the same issue
Same goes for Azure functions. Getting the message:

> Error: uuid: This browser does not seem to support crypto.getRandomValues(). If you need to support this browser, please provide a custom random number generator through options.rng.

ctavan:
@evdama @jvpelt @VandalPaulius @ZebraFlesh could you please share more details about your Azure Functions/AWS Lambda/Google Cloud Functions setup with steps to reproduce the error?

My suspicion is that you are bundling code in a way that is designed to run in a browser environment, not in Node.js

So are any bundlers like webpack or rollup involved in your process?

I created a minimalistic PoC that shows that under normal circumstances using `uuid` in Cloud Functions works just well: https://github.com/ctavan/uuid-cloud-functions-poc

evdama:
yes, I use rollup to bundle uuid

ZebraFlesh:
@ctavan Here's a minimal repo with what I'm doing. master branch is set to uuid@3.4.0; upgrade to uuid@7.0.0 to see failure. https://github.com/ZebraFlesh/uuid-serverless-poc

ctavan:
Thank you @ZebraFlesh. I think a fix is near: #380

VandalPaulius:
> 
> So are any bundlers like webpack or rollup involved in your process?
> 
> I created a minimalistic PoC that shows that under normal circumstances using `uuid` in Cloud Functions works just well: https://github.com/ctavan/uuid-cloud-functions-poc

Yes, in my case I was using Webpack to bundle everything, but the target was set to `node`. Hopefully #380 will fix it

ctavan:
I have just released `uuid@7.0.1` which should fix this issue. If the problem remains for anyone please feel free to open a new issue.

And thanks @TrySound for the great work! 🎉

evdama:
Thanks much, working again for firebase cloud functions with node express middleware (code see above).

ZebraFlesh:
Thanks so much for the quick turn around! Confirmed on my end that my PoC is fixed with 7.0.1 🙌

[...]

PR #380, opened 2020-02-25
fix: provide browser versions independent from module system

TrySound:
Probably will fix https://github.com/uuidjs/uuid/issues/378

Currently main field is used for node and module for browser build. This
is not entierly correct. Webpack can be used to build node apps. Though
module is always preferred over main. For browser versions there is a
special field.

In this diff I added esm support for both node and browser.
rollup-plugin-node-resolve look at browser field as well and prefer it
to bundle umd.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,13 @@
   },
   "sideEffects": false,
   "main": "dist/index.js",
-  "module": "dist/esm-browser/index.js",
+  "module": "dist/esm-node/index.js",
+  "browser": {
+    "./dist/md5.js": "./dist/md5-browser.js",
+    "./dist/rng.js": "./dist/rng-browser.js",
+    "./dist/sha1.js": "./dist/sha1-browser.js",
+    "./dist/esm-node/index.js": "./dist/esm-browser/index.js"
+  },
   "files": [
     "CHANGELOG.md",
     "CONTRIBUTING.md",
@@ -36,6 +42,7 @@
     "@babel/preset-env": "7.8.4",
     "@commitlint/cli": "8.3.5",
     "@commitlint/config-conventional": "8.3.4",
+    "@rollup/plugin-node-resolve": "7.1.1",
     "babel-eslint": "10.0.3",
     "babel-plugin-add-module-exports": "1.0.2",
     "browserstack-local": "1.4.5",
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -14,12 +14,17 @@ mkdir -p "$DIR"
 # Transpile CommonJS versions of files
 babel --env-name commonjs src --source-root src --out-dir "$DIR" --copy-files --quiet
 
-# Transpile ESM versions of files for the browser
+# Transpile ESM versions of files for browser
 babel --env-name esm src --source-root src --out-dir "$DIR/esm-browser" --copy-files --quiet
+
+# Transpile ESM versions of files for node
+babel --env-name esm src --source-root src --out-dir "$DIR/esm-node" --copy-files --quiet
 
 # No need to have the CLI files in the esm build
 rm -rf "$DIR/esm-browser/bin"
 rm -rf "$DIR/esm-browser/uuid-bin.js"
+rm -rf "$DIR/esm-node/bin"
+rm -rf "$DIR/esm-node/uuid-bin.js"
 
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -18,7 +18,15 @@
   },
   "sideEffects": false,
   "main": "dist/index.js",
-  "module": "dist/esm-browser/index.js",
+  "module": "dist/esm/index.js",
```
ctavan: I'm wondering if this field now still makes sense? If I read https://webpack.js.org/configuration/resolve/#resolvemainfields correctly then, webpack will look for

- `['browser', 'module', 'main']` (in that order) when building for the browser.
- `['module', 'main']` when building for node.

Does that mean webpack will use the `module` entrypoint of the code when bundling for Node.js and transform everything into CommonJS for Node.js?

Do we agree that this has not yet anything to do with native ESM support for node?

scripts/build.sh
```diff
@@ -14,21 +14,12 @@ mkdir -p "$DIR"
 # Transpile CommonJS versions of files
 babel --env-name commonjs src --source-root src --out-dir "$DIR" --copy-files --quiet
 
-# Transpile ESM versions of files for the browser
-babel --env-name esm src --source-root src --out-dir "$DIR/esm-browser" --copy-files --quiet
+# Transpile ESM versions of files
```
ctavan: `Transpile ESM versions of files for use with module bundlers` as we do not yet support native Node.js esmodules.

ctavan:
Also, please rebase as master has changed meanwhile.

package.json
TrySound: > Does that mean webpack will use the module entrypoint of the code when bundling for Node.js and transform everything into CommonJS for Node.js?

It does not transform into commonjs. ESM is used for more efficient bundling into one file. The result bundle can be any format configured in webpack (esm too since webpack v5)

> Do we agree that this has not yet anything to do with native ESM support for node?
No "module" field is bundler specific tool

scripts/build.sh
TrySound: We should support bundlers with node target.

package.json
ctavan: So yes, we do agree that this is something bundler specific then 😉 ?!

ctavan: So yes, we do agree that this is something bundler specific then 😉 ?!

scripts/build.sh
ctavan: The point that I wanted to make is that this is still designated for module bundlers and not https://nodejs.org/api/esm.html

package.json
TrySound: Yes, both module and browser fields are bundler specific.

scripts/build.sh
ctavan: So: I agree we should not restrict this to browser but the esm build here is still targeted at module bundlers!

TrySound: Yes, though commonjs browser stuff can be used by browserify unless it supports modules too.

TrySound:
@ctavan Added links to commit message. Tests look unstable, not related to my changes.

ctavan:
@TrySound the code looks good, the build errors are due to lacking Browserstack credentials from your fork. I still have to figure out how to proceed with that.

TrySound:
You mean changed paths is breaking change?

TrySound:
Ah, got it. Browsers do not get right paths.

TrySound:
Ok, I can do it.

ctavan:
No: The thing is that so far we only had *one* esm build designated at the browser where the browser-specific files have been renamed to just replace the node-specific files. That way, you could use the ESM build directly in the browser.

With your change, you rely on the `browser` field in `package.json` which is respected by all relevant bundlers but browsers with native esmodules support don't care about that.

So unfortunately your change breaks the opportunity to use esmodules directly in the browser as shown in the example: https://github.com/uuidjs/uuid/blob/master/examples/browser-esmodules/

Since we don't have the nice workaround of relying on the `browsers` field we may need to build two separate esmodule variants: one that can natively be consumed by the browser and one that can natively be consumed by node. We can then still make use of the `browser` field in package.json in order to point module bundlers to the correct files for all environments.

TrySound:
@ctavan Looks good now?

Issue #392, opened 2020-03-02
getRandomValues couldn't be found from a WebWorker in Edge18


evertbouw:
## Describe the bug

I've used uuid v4 in a webworker but now I'm seeing error logs from Edge18.

```
SCRIPT0: SCRIPT0: uuid: This browser does not seem to support crypto.getRandomValues(). If you need to support this browser, please provide a custom random number generator through options.rng.
27788d0bfdaa0d0b9446.worker.js (2,1)
```

## Runtime

- OS: Windows 10
- Runtime: Edge
- Runtime Version: EdgeHTML 18.19041

ctavan:
@evertbouw can you please provide steps to reproduce as requested by the issue template? Thanks!

evertbouw:
I thought that would look a little bit silly as it's just repeating `call v4 in a webworker in Edge`, but I'll try to get a minimal repro online

evertbouw:
![image](https://user-images.githubusercontent.com/6398211/75704929-90b9f580-5cba-11ea-8287-202841294482.png)

live: https://evertbouw.github.io/buggy-edge/
code: https://github.com/evertbouw/buggy-edge/blob/master/src/foo.worker.js

ctavan:
Thank you for the reproduction code, that was really helpful!

Unfortunately, according to https://caniuse.com/#feat=cryptography, the WebCrypto API is not supported in WebWorkers in Edge<=18.

There was an issue regarding that problem on the old Edge Issuetracker which is no longer available: https://web.archive.org/web/20171210184325/https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7607496/

Unfortunately this means that if you want to continue to support pre-Chromium Edge Versions you will either have to continue to use `uuid@3.4.0` or you have to supply your own random number generator to the `v4()` method [through `options.rng`](https://github.com/uuidjs/uuid#version-4-random).

You could take a look at the old version of this library where we were shipping an insecure fallback random number generator for environments that do not support proper cryptographically secure random number generators: https://github.com/uuidjs/uuid/blob/v3.4.0/lib/rng-browser.js#L24-L33

However, please be aware that using a flawed random number generator might lead to unexpected consequences, see https://github.com/tc39/proposal-uuid#how-unique-are-v4-uuids for examples.

broofa:
[...]

PR #395, opened 2020-03-03
chore: add webworker to not supported table 

broofa:
For #392

### DIFF
--- a/README_js.md
+++ b/README_js.md
@@ -340,6 +340,12 @@ import 'react-native-get-random-values';
 import { v4 as uuidv4 } from 'uuid';
 ```
 
+### Web Workers / Service Workers (Edge <= 18)
+
+[In Edge <= 18, Web Crypto is not supported in Web Workers or Service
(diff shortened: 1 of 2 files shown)


README_js.md
```diff
@@ -333,6 +333,7 @@ API is not supported. This issue can be resolved by adding an appropriate polyfi
 | Environment  | Suggested Polyfill                                                                                  |
 | ------------ | --------------------------------------------------------------------------------------------------- |
 | React Native | [`react-native-get-random-values`](https://github.com/LinusU/react-native-get-random-values#readme) |
+| WebWorker (Edge) | N/A (let us know if you find something) |
```
TrySound: Worth to mention only old edge have this problem

broofa: Sure... do we have data on which Edge versions do/don't support crypto in webworkers?  The only thing I've found is the [caniuse page](https://caniuse.com/#feat=cryptography), which doesn't go into detail on versions.

TrySound: It does. 18 and below is old edge. 79+ is chromium based
image

ctavan:
@broofa adjusted from the table to section layout.

PR #402, opened 2020-03-07
feat: native Node.js ES Modules (isolated state)

### LINKED ISSUE #342: ECMAScript Module for Node.js
As of 2019-10-31 ECMAScript Module (ESM) support in Node.js is still experimental and there is not yet a stable way of offering an ESM build for Node.js. The authors of the Node.js module working group even urge developers not to release dual CommonJS/ESM packages to npm yet, see: https://github.com/nodejs/modules/blob/master/doc/plan-for-new-modules-implementation.md#phase-4-further-improvements-after-unflagging which states:

> * Dual CommonJS/ESM packages: Either support packages with both CommonJS and ESM sources that can be used in either environment; or decide to specifically not support dual CommonJS/ESM packages.
>   * Status quo (at time of possible unflagging): "main" points to exactly one file, and all file extensions are mandatory (by default), so there is no possibility of an import specifier pointing to different files in ESM versus CommonJS. Recommended practice for dual packages is to have "main" point to the CommonJS entry point and have users use a deep import, e.g. /module.mjs, to access ESM entry point.

There is ongoing work within Node.js to support multiple exports for npm packages which will eventually solve this issue. We will support this feature once it becomes stable in node (i.e. once it's no longer hidden behind an experimental flag).

Corresponding Node.js issues:
 * https://github.com/nodejs/node/pull/29978

ctavan:
See: https://nodejs.org/docs/latest-v12.x/api/esm.html#esm_dual_commonjs_es_module_packages

Potentially closes #342

----

The native esmodule for node.js now uses an `.mjs` wrapper around the transpiled CommonJS code in order to avoid the [dual package hazard](https://nodejs.org/docs/latest-v12.x/api/esm.html#esm_dual_package_hazard) which potentially affects `v1` uuid creation since we store internal state in the module: https://github.com/uuidjs/uuid/blob/ee039eed77dc2bc77244e641c1a4823df2b1a2dc/src/v1.js#L9-L14

I'm not yet sure how this wrapper approach will interact with the node-specific esm build that we currently generate for module bundlers like webpack/rollup (for use in serverless environments): https://github.com/uuidjs/uuid/blob/ee039eed77dc2bc77244e641c1a4823df2b1a2dc/scripts/build.sh#L21

And again this whole stuff is still experimental in Node.js, so maybe it's too early to land this.

- [x] Clarify interaction with `esm-node` build.
- [x] Add documentation to README
- [x] Add test case
- [x] Clarify the need for rollup `commonjs()` plugin -> https://github.com/rollup/rollup/issues/3514
  - Rollup considers itself as a pure ESM bundler, so supporting CommonJS modules out of the box seems out of scope for them. This means that we cannot ship the "isolated state" code for bundling with rollup. The wrapper approach proposed in #423 therefore seems to be the more robust solution.


package.json
```diff
@@ -18,6 +18,10 @@
   },
   "sideEffects": false,
   "main": "dist/index.js",
+  "exports": {
+    "require": "dist/index.js",
```
TrySound: require will trigger warnings on old node 13 versions. Looks like there is a solution around this
https://twitter.com/lukastaegert/status/1236188083621687298

TrySound: Ah, here's what rollup end up with. Probably no need in mjs wrapper which would not work when exports support will come to bundlers.
https://github.com/rollup/rollup/pull/3430/files

ctavan: Thanks for the hint. Although I’m definitely not planning to add workarounds for experimental APIs of old non-LTS node versions.

ctavan: Also: how does the rollup solution prevent the dual package hazard?

TrySound: I'm not sure. Looks like some fields are just ignored in early node 13.
https://twitter.com/guybedford/status/1235329049779634177
Or what do you mean by dual package hazard?

ctavan: https://nodejs.org/docs/latest/api/esm.html#esm_dual_package_hazard

And if you want to dig deeper: https://github.com/nodejs/modules/issues/371

TrySound: uuid is stateless package. A few instances should not be a problem. It's the same as a few versions.

broofa:
> uuid is stateless package.

** cough ** [stateless](https://github.com/uuidjs/uuid/blob/master/src/v1.js#L9-L14) ** cough **

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F80TEu4wOBdPLG%2Fgiphy.gif&token=oops" data-gitmeme-token="oops"></a>

package.json
broofa: > uuid is stateless package.

** cough ** [stateless](https://github.com/uuidjs/uuid/blob/master/src/v1.js#L9-L14) ** cough **

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F80TEu4wOBdPLG%2Fgiphy.gif&token=oops" data-gitmeme-token="oops"></a>

TrySound: is it not?

broofa: Well... it doesn't expose state externally, but v1 has those private state vars it maintains. For functional programming purists and anyone looking for idempotent behavior, v1 isn't gonna cut it.

TrySound: got it. thanks.
well, its on dependent packages to provide es modules support. we can only ask to open issues on those packages instead.

ctavan: I think there are 2 aspects w.r.t. uuid and the dual package hazard:

1. I agree that most likely in practice it wouldn't be a problem to just not care about the `v1()` internal state…
1. On the other hand, `uuid` is so heavily depended upon that it is probably one of the very few modules on npm where the dual package hazard will actually manifest in the real world! I just checked a couple of normal web projects and found that `uuid` often appears up to 10-20 times as a transitive dependency: So unless the entire dependency chain was either 100% ESModule or 100% CommonJS (and that will likely not be the case for a very long time) the dual package hazard will manifest for real!

Again, I think it's unlikely that this will cause bugs in the real world, but it still somewhat hurts the pedantic and principled software engineering regions of my brains 😂

TrySound: Funny thing is that uuid in transitive dependencies may be v3 for ages which is not much different from dual packages hazard and definitely not our responsibility. It's also possible the most used api is stateless v4.

ctavan: That’s a very valid point and in practice probably much more important than my musings about the dual package hazard, I totally agree with you. 

But that still doesn’t convince me that we shouldn’t care about it and at least be exactly aware of what we are doing 😉

ctavan:
FYI I have built reproduction examples for the dual package hazard: https://github.com/ctavan/uuid-dual-package-hazard

The two solutions described in https://nodejs.org/api/esm.html#esm_dual_commonjs_es_module_packages (wrapper & state isolation) work well.

I tend towards using state isolation since this would allow us to ship _one_ ESM build for node that can both directly be consumed by modern node versions as well as by module bundlers when targeting node.

If we use the wrapper technique then I think we'll have to ship a separate ESM node build for bundlers. Otherwise bundlers would lose the ability to tree-shake Node.js code (which might not be a big deal, but also unnecessary).

Opinions?

src/v1state.cjs
```diff
@@ -0,0 +1,7 @@
+module.exports = exports.default = {
```
TrySound: module.exports.default. exports has old reference.

examples/browser-rollup/package.json
```diff
@@ -10,6 +10,7 @@
     "uuid": "file:../../.local"
   },
   "devDependencies": {
+    "@rollup/plugin-commonjs": "^11.0.2",
     "rollup": "^1.32.0",
     "rollup-plugin-node-resolve": "^5.2.0",
```
TrySound: `@rollup/plugin-node-resolve`

ctavan: @TrySound when I upgrade to `@rollup/plugin-node-resolve` the `{browser: true}` setting seems to be ignored. Here's the output of an `npm run build` inside `examples/browser-rollup`:

```
./example-all.js → dist/all.js...
(!) Plugin node-resolve: preferring built-in module 'crypto' over local alternative at 'crypto', pass 'preferBuiltins: false' to disable this behavior or 'preferBuiltins: true' to disable this warning
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
crypto (imported by ../../dist/esm-node/sha1.js, ../../dist/esm-node/rng.js, ../../dist/esm-node/md5.js)
(!) Missing global variable name
Use output.globals to specify browser global variable names corresponding to external modules
crypto (guessing 'crypto')
created dist/all.js in 417ms
```

Any idea why?

TrySound: After an hour of debugging I figured out the problem is symlink to uuid. The new version does not handle it properly.

ctavan: @TrySound thanks for looking into this, you are of great help for this library!

Have you been able to figure out if that is an intended behavior of the new rollup plugin or a bug? If so, would you mind reporting it upstream with the information you discovered?

TrySound: I think it's a bug. I'll try to write test and fix the issue. Keep old package for now.

TrySound: By the way "preserveSymlinks: true" in rollup config solves the problem with new package.

TrySound: Fixed here https://github.com/rollup/plugins/pull/291

frank-dspeed:
Hello I am Frank Lemanschik a German Senior ECMAScript Consultant from the rollup Team i will investigate and clone https://github.com/uuidjs/uuid/tree/node-esm

 I will do all the needed and review it with the newest nodejs and older nodejs versions I have researched the best packaging patterns.

frank-dspeed:
@ctavan msged you on gitter maybe we can solve that more stable and fast when we talk before as I think there are some other changes needed maybe also I am not sure,

ctavan:
Rollup considers itself as a pure ESM bundler, so supporting CommonJS modules out of the box seems out of scope for them. This means that we cannot ship the "isolated state" code for bundling with rollup. The wrapper approach proposed in #423 therefore seems to be the more robust solution.

PR #411, opened 2020-03-31
feat: drop Node.js 8.x support

### LINKED ISSUE #410: Drop Node.js 8.x support
Node.js 8.x [reached EOL on 2019-12-31](https://github.com/nodejs/Release#end-of-life-releases).

Many of our development dependencies (prettier, rollup) already removed support for Node 8.x.

I would like to remove official support for node 8.x as well. This will require a major version bump.

Thoughts?

ctavan:
BREAKING CHANGE: Drop official support for Node.js 8.x since the 8.x
line has reached end-of-life on 2019-12-31. The library is likely to
continue to work with Node.js 8.x, we simply won't provide official
support for it.

Fixes #410.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x]
+        node-version: [10.x, 12.x]
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


ctavan:
> "You've been warned", I guess? :-)

Hmmmm 🤔 

Runtimes like [Google Cloud Functions](https://cloud.google.com/functions/docs/concepts/nodejs-8-runtime) still support Node.js 8.x…

The only reason I want to drop Node.js 8.x support for now is to be able to use tooling that now requires Node.js 10.x…

An alternative would be to find a way to build the code using e.g. Node.js 12.x but then run the test suite using Node.js 8.x… I'll give that a shot first.

ctavan:
@broofa see #413 for an alternative approach. I think I prefer #413 for now.

ctavan:
Dropped for now in favor of #413.

PR #412, opened 2020-03-31
Upgrade dependencies

ctavan:
Now that we run certain parts of the CI and build process only in the latest node release (#415) we can upgrade our toolchain which removed support for Node.js 8.x in many libraries.


TrySound:
New node-resolve version is released

src/md5.js
```diff
@@ -7,10 +7,7 @@ function md5(bytes) {
     bytes = Buffer.from(bytes, 'utf8');
   }
 
-  return crypto
-    .createHash('md5')
-    .update(bytes)
-    .digest();
+  return crypto.createHash('md5').update(bytes).digest();
```
LinusU, line 10: 😍

ctavan: Well, that was the prettier-update…

PR #426, opened 2020-04-29
feat: remove deep requires

ctavan:
BREAKING CHANGE: Deep requiring specific algorithms of this library like
require('uuid/v4'), which has been deprecated in uuid@7, is no longer
supported.

Instead use the named exports that this module exports.

For ECMAScript Modules (ESM):

```javascript
uuidv4();
```

For CommonJS:

```javascript
const { v4: uuidv4 } = require('uuid');
uuidv4();
```

No longer supported is this:

```javascript
const uuidv4 = require('uuid/v4'); // <== NO LONGER SUPPORTED!
uuidv4();
```


ctavan:
I'll wait for @broofa's final go before merging this and pulling out a new major version release.

PR #427, opened 2020-04-29
refactor: upgraded code to es6 standard

awwit:
Changes:
Improved eslint configuration.
Upgraded code to es6 standard.
Optimized use of arrays.

Since you are already using ES6 features (`Uint8Array` as example), I allowed myself to adapt your code to this standard.

All `var` were replaced by `let` or `const`. So we will give more optimization opportunities for JS engine.

The same goes for using arrays. It is not recommended to create arrays through `new Array(N)`. Because arrays are created full of holes. Which affects performance.
https://v8.dev/blog/elements-kinds

Since you are already using `Uint8Array`, I replaced regular arrays with typed ones (in some places). Which are created with a fixed size and free from the disadvantages of regular arrays.

P.S. If something does not suit you in my changes, then let me know and I will fix it =)


src/v1.js
```diff
@@ -48,7 +48,7 @@ function v1(options, buf, offset) {
   // (1582-10-15 00:00).  JSNumbers aren't precise enough for this, so
   // time is handled internally as 'msecs' (integer milliseconds) and 'nsecs'
   // (100-nanoseconds offset from msecs) since unix epoch, 1970-01-01 00:00.
-  let msecs = options.msecs !== undefined ? options.msecs : Date.now();
+  let msecs = options.msecs !== undefined ? options.msecs : new Date().getTime();
```
LinusU, line 51: Just out of curiosity, why was `new Date().getTime()` preferred over `Date.now()`? ☺️

package.json
```diff
@@ -77,19 +82,19 @@
     "examples:node:commonjs:test": "cd examples/node-commonjs && npm install && npm test",
     "examples:node:esmodules:test": "cd examples/node-esmodules && npm install && npm test",
     "lint": "npm run eslint:check && npm run prettier:check",
-    "eslint:check": "eslint src/ test/ examples/ *.js",
-    "eslint:fix": "eslint --fix src/ test/ examples/ *.js",
+    "eslint:check": "npx eslint src/ test/ examples/ *.js",
```
LinusU, line 85: Both npm and yarn makes sure that `node_modules/.bin` is first in the path, is there another reason to add `npx` here? 🤔

src/v1.js
awwit: Because with `Date.now()` the test `docs:diff` fails… >_<

.eslintrc.json
```diff
@@ -7,7 +7,14 @@
...
     "node": true
   },
-  "extends": ["eslint:recommended", "plugin:prettier/recommended"],
+  "plugins": ["node", "standard", "prettier"],
+  "extends": [
+    "eslint:recommended",
+    "standard",
+    "prettier",
+    "prettier/standard",
+    "plugin:prettier/recommended"
```
TrySound, line 16: Why not just run prettier --list-different?

package.json
awwit: Just in case. It happens that the modules are not fully installed and then their global versions can be called.

.eslintrc.json
awwit: By connecting the prettier to eslint, we can automatically correct the errors of the prettier (through eslint).
or what did you mean?

src/v1.js
ctavan: For updating the docs please edit `README_js.md` and then `npm run docs`, then commit the changes to both `README_js.md` and `README.md`.

awwit:
@ctavan if you want me to make a benchmark, then I will do it. But a little later.

What test tools do you prefer? or is it not that important?

ctavan:
> @ctavan if you want me to make a benchmark, then I will do it. But a little later.
> 
> What test tools do you prefer? or is it not that important?

Great! I haven't done performance testing of JavaScript libraries recently so I would have to investigate what the state-of-the art is.

The only thing that I would consider important is that performance can be tested both, in Node.js as well as in browsers.

The performance tests should probably also be committed to this repository.

src/v1.js
awwit: @ctavan with `Date.now()` different values are always generated. Therefore it is impossible to pass the test `docs:diff`. Try it yourself. I spent an hour to understand what was happening )

.eslintrc.json
TrySound: IMO eslint should not be used for formatting at all

awwit: Why not? The project already uses `eslint --fix` and `prettier --write` (in lint-staged).
It turns out that the same files are being processed twice (by eslint and prettier).
You can process them in one pass, if you connect the prettier inside eslint.

In addition, if you use them separately, their rules may conflict.
Then in a text editor they constantly correct each other (if plugins are installed for eslint and prettier).

src/bytesToUuid.js
```diff
@@ -2,36 +2,39 @@
...
+const byteToHex = [];
+
+for (let i = 0; i < 256; ++i) {
+  byteToHex.push((i + 0x100).toString(16).substr(1));
 }
 
 function bytesToUuid(buf, offset) {
-  var i = offset || 0;
-  var bth = byteToHex;
+  const i = offset || 0;
```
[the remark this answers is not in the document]
awwit: Yes, IE 11 does not support this feature:
http://kangax.github.io/compat-table/es6/#test-default_function_params

broofa:
> exploring ways of benchmarking the library

For the record, I used to track performance in this module  (using [ad-hoc perf tests](https://github.com/uuidjs/uuid/blob/v1.4.7/test/test.js#L217-L231)).  I removed that code because perf simply stopped being interesting.   It's been a while but last I checked I think the #'s were on the order of 100K's or even 1M's of ids/second (for v1 and v4).

I eventually concluded that the overhead of UUID creation almost always paled in comparison to whatever else the user was doing with that UUID (e.g. creating a DB entry or creating/validating model data).

That said, having a performance test as part of our CI tests to catch order-of-magnitude regressions would be good, but I don't think we need it as part of this PR.  Let's create a separate issue for that.

ctavan:
@broofa I totally agree that performance is currently most likely not a problem of the `uuid` library. All I want is to catch, as you write, "order-of-magnitude regressions".

I've added a little script that might already do the trick: #430 

Would be curious to see it run before and after this PR. @awwit wanna give it a try?

awwit:
@ctavan I also already managed to prepare the code for performance testing (also using `benchmark`).

Almost the same code turned out as yours. I have to try it.

(In my version, I output the results to a file `${git-branch}-${node--version}.json`. So you can save test results and switch to other branches.)

Let's try your benchmark. You need to merge it into a master. Then I can merge it into my branch and see the difference in performance.

ctavan:
@awwit I've merged the benchmark script to master. Could you rebase this PR so we can see the results?

Thanks!

ctavan:
> My one concern is that embracing ES6 is going to encourage the use of patterns and features that may break IE11, [where there's limited ES6 support](http://kangax.github.io/compat-table/es6/#ie11). E.g. see my comment below about using a default value for the `offset` argument. Such bugs should get caught by our BrowserStack tests though (right, @ctavan?)

Regarding JS compatibility: Since we [use babel to transpile the source](https://github.com/uuidjs/uuid/blob/2b796869ec04b5f8df3c419ef0d9d76e4f6925f0/scripts/build.sh#L14-L21) into a form to support all platforms that we want to support we should not need to worry too much.

Also our tests should catch regressions for all currently supported platforms 🤞 ...

The only thing we should be careful about is that the more we let babel transpile code, the higher the risk that we get

1. increased bundlesize due to poor transformations of babel (we have bundlewatch to catch that!)
1. performance regressions (we have the benchmark script now to catch that!)

.eslintrc.json
ctavan, line 16: `"extends": ["plugin:prettier/recommended"]` already takes care of loading the `prettier` eslint plugin, so it should not be necessary to specify it again in the `plugins` array, see: https://prettier.io/docs/en/integrating-with-linters.html#recommended-configuration

src/md5-browser.js
```diff
@@ -19,10 +19,14 @@
...
  */
 function md5(bytes) {
-  if (typeof bytes == 'string') {
-    var msg = unescape(encodeURIComponent(bytes)); // UTF8 escape
-    bytes = new Array(msg.length);
-    for (var i = 0; i < msg.length; i++) bytes[i] = msg.charCodeAt(i);
+  if (typeof bytes === 'string') {
+    const msg = unescape(encodeURIComponent(bytes)); // UTF8 escape
+
+    bytes = new Uint8Array(msg.length);
```
ctavan, line 25: Why a Uint8Array instead of a plain Array?

src/md5-browser.js
```diff
@@ -155,14 +159,15 @@ function wordsToMd5(x, len) {
...
-    output[i] = 0;
+  if (input.length === 0) {
+    return [];
   }
-  var length8 = input.length * 8;
-  for (i = 0; i < length8; i += 8) {
+
+  const length8 = input.length * 8;
+
+  const output = new Uint32Array(getOutputLength(length8));
```
ctavan, line 168: Why a `Uint32Array` instead of a plain `Array`?

src/rng-browser.js
```diff
@@ -4,18 +4,22 @@
...
 
-var rnds8 = new Uint8Array(16); // eslint-disable-line no-undef
 export default function rng() {
   if (!getRandomValues) {
     throw new Error(
       'crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported',
     );
   }
+
+  const rnds8 = new Uint8Array(16);
```
ctavan, line 22: This might have a performance impact. A new `Uint8Array` is created with each call to `rng()` instead of reusing a preallocated array. Please verify!

src/sha1-browser.js
```diff
@@ -18,53 +18,63 @@ function ROTL(x, n) {
...
 
-  var l = bytes.length / 4 + 2;
-  var N = Math.ceil(l / 16);
-  var M = new Array(N);
+  const l = bytes.length / 4 + 2;
+  const N = Math.ceil(l / 16);
+  const M = new Array(N);
+
+  for (let i = 0; i < N; ++i) {
+    const arr = new Uint32Array(16);
```
ctavan, line 41: Same here: Let's verify if we can use typed arrays without performance penalty.

src/sha1-browser.js
```diff
@@ -18,53 +18,63 @@ function ROTL(x, n) {
...
   }
 
   M[N - 1][14] = ((bytes.length - 1) * 8) / Math.pow(2, 32);
   M[N - 1][14] = Math.floor(M[N - 1][14]);
   M[N - 1][15] = ((bytes.length - 1) * 8) & 0xffffffff;
 
-  for (var i = 0; i < N; i++) {
-    var W = new Array(80);
+  for (let i = 0; i < N; ++i) {
+    const W = new Uint32Array(80);
```
ctavan, line 59: See above.

src/v4.js
```diff
@@ -2,23 +2,24 @@ import rng from './rng.js';
 import bytesToUuid from './bytesToUuid.js';
 
 function v4(options, buf, offset) {
-  var i = (buf && offset) || 0;
+  const i = (buf && offset) || 0;
 
-  if (typeof options == 'string') {
-    buf = options === 'binary' ? new Array(16) : null;
+  if (typeof options === 'string') {
+    buf = options === 'binary' ? new Uint32Array(16) : null;
```
ctavan, line 8: See above re typed arrays.

awwit:
@ctavan done, I merged your tests to my branch.
Now we can look at the performance.

.eslintrc.json
```diff
@@ -7,7 +7,14 @@
     "jest": true,
     "node": true
   },
-  "extends": ["eslint:recommended", "plugin:prettier/recommended"],
+  "plugins": ["node", "standard", "prettier"],
+  "extends": [
+    "eslint:recommended",
+    "standard",
```
ctavan, line 13: If we decide to go for a standard eslint preset, I think we should try to get rid of as many custom rules as possible. I would be all in for that!

How did you decide which custom rules to keep and which to drop? Are there maybe more rules we can drop?

ctavan:
Node 12

Master:

```
uuidv1() x 1,319,851 ops/sec ±1.76% (86 runs sampled)
uuidv1() fill existing array x 4,732,686 ops/sec ±1.83% (90 runs sampled)
uuidv4() x 294,904 ops/sec ±2.72% (82 runs sampled)
uuidv4() fill existing array x 357,955 ops/sec ±3.05% (82 runs sampled)
uuidv3() x 106,752 ops/sec ±3.61% (79 runs sampled)
uuidv5() x 113,064 ops/sec ±1.56% (83 runs sampled)
Fastest is uuidv1() fill existing array
```

This branch:

```
uuidv1() x 1,519,697 ops/sec ±2.50% (87 runs sampled)
uuidv1() fill existing array x 8,119,318 ops/sec ±2.89% (92 runs sampled)
uuidv4() x 303,734 ops/sec ±2.53% (86 runs sampled)
uuidv4() fill existing array x 362,363 ops/sec ±4.08% (85 runs sampled)
uuidv3() x 114,053 ops/sec ±1.37% (82 runs sampled)
uuidv5() x 108,882 ops/sec ±2.21% (80 runs sampled)
Fastest is uuidv1() fill existing array
```

`uuidv1()` performance has improved, especially in the "fill existing array" case 😲 ! Would be curious to learn what caused this improvement.

The rest seems to perform similar than before (within error margins) 👏

src/rng-browser.js
awwit: I think this is unsafe, because someone can use your `rng()` function in their asynchronous code. When generating random numbers, it will always receive the same array. (New data will overwrite old).

Multiple function calls will return the same array.

src/md5-browser.js
awwit: In general, typed arrays should work faster, since the JS engine will immediately understand what data it works with.

https://bugs.chromium.org/p/v8/issues/detail?id=2190

You can see the tests at this link ^

.eslintrc.json
awwit: Ok, I'll see what rules can be safely removed.

src/rng-browser.js
broofa: `rng-browser` is not part of the public API, so should not be used externally.  Internally, there's no async code that uses it.  It's also very purpose-specific (generating a 16-byte array of random #'s), so I don't see there being a lot of use cases where someone would actually need to figure out how to reach into this module to access this function.  Thus, the risk to users feels more hypothetical than real.

That said, reusing the array was an optimization I put in place [a long time ago](https://github.com/uuidjs/uuid/commit/774c6e3169dd04be2a80af14153d9379586dd064#diff-bef26ab2ce6b354722523f65bc301d59R16).  There was no real demand for it (nobody was complaining about perf at the time).  It was just a vanity thing to see how much perf I could squeeze out of the code.

Now, _having said that_, it's looks like it's still a hell of an optimization, which I'll suggest is worth keeping.

[/me pats himself on back 😆]

On `master`:
```
rng() reusing array:
uuidv4() x 507,182 ops/sec ±5.43% (63 runs sampled)
benchmark.js:32 uuidv4() fill existing array x 688,903 ops/sec ±8.40% (63 runs sampled)

rng() creating new array each time:
uuidv4() x 183,198 ops/sec ±10.90% (45 runs sampled)
benchmark.js:32 uuidv4() fill existing array x 211,791 ops/sec ±13.46% (41 runs sampled)
```

On this PR:
```
rng() reusing array:
uuidv4() x 524,604 ops/sec ±0.82% (64 runs sampled)
benchmark.js:20 uuidv4() fill existing array x 741,362 ops/sec ±0.52% (64 runs sampled)

rng() creating new array each time:
uuidv4() x 197,603 ops/sec ±10.09% (51 runs sampled)
benchmark.js:20 uuidv4() fill existing array x 211,450 ops/sec ±15.85% (41 runs sampled)
```

(Note: I suspect the larger margin of error in the new-array case is due is because the increased level of array creation results in more erratic garbage collection.)

src/bytesToUuid.js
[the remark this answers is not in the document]
broofa: Since @ctavan says babel compilation should take care of ES6 issues for IE11, and our browser stack tests should catch the issue if not, let's go ahead and make this change.

src/md5-browser.js
awwit: https://v8.dev/blog/elements-kinds

src/md5-browser.js
awwit: https://v8.dev/blog/elements-kinds

src/md5-browser.js
broofa: Nice.  When I first added v3/v5 support, I wanted to insure support for older, legacy platforms that didn't have typed array support.  That's no longer a concern.

src/md5-browser.js
broofa: Should allow for better JS engine optimizations.  'Pretty sure typed arrays are supported on all our supported platforms, so this should be okay.

broofa:
> Thanks for your effort! ... this is a valuable contribution and will help us modernize the code base. ... I think in general this already looks really good!

Agreed!

> `uuidv1()` performance has improved, especially in the "fill existing array" case 😲 ! Would be curious to learn what caused this improvement.

Me too.  I know switching from `var` to `const`/`let` is generally considered a good thing (perf-wise).  As is using typed arrays, but I'd be curious to hear what someone familiar with how JS engines work would have to say about this.

awwit:
@ctavan 

> `uuidv1()` performance has improved, especially in the "fill existing array" case  ! Would be curious to learn what caused this improvement.

I checked, this is due to the fact that we replaced `new Date().getTime()` to `Date.now()`.

src/rng-browser.js
ctavan: > [/me pats himself on back 😆]

Well played, 9-year-ago-you, well played! 😂

src/md5-browser.js
ctavan: Hmm, what made me suspicious is the following. With this adjustment to the benchmarks:

```diff
--- a/examples/benchmark/benchmark.js
+++ b/examples/benchmark/benchmark.js
@@ -7,7 +7,7 @@ const uuidv5 = (typeof window !== 'undefined' && window.uuidv5) || require('uuid

 console.log('Starting. Tests take ~1 minute to run ...');

-const array = new Array(16);
+const array = new Uint8Array(16);
 const suite = new Benchmark.Suite();
 suite
   .add('uuidv1()', function () {
```

the results change significantly.

With `new Array(16)`:

```
uuidv1() fill existing array x 4,749,694 ops/sec ±3.14% (87 runs sampled)
uuidv4() fill existing array x 356,833 ops/sec ±3.75% (83 runs sampled)
```

With `new Uint8Array(16)`:

```
uuidv1() fill existing array x 775,015 ops/sec ±1.50% (93 runs sampled)
uuidv4() fill existing array x 364,461 ops/sec ±4.00% (85 runs sampled)
```

What am I missing?

src/rng-browser.js
ctavan: @awwit please re-introduce this optimization by moving the array definition out of the function scope again.

src/md5-browser.js
awwit: @ctavan hmm…

On my machine the difference in the margin of error. Try again…
Although, I made some more optimizations locally.

awwit:
@ctavan Bundlewatch test failed =/

Screenshot_20200502_001543

We will have to increase the size to get more optimizations (especially parsing uuid bytes).

If the size is so critical, then I can remove typed arrays in failed bundles.

src/bytesToUuid.js
```diff
@@ -14,28 +14,28 @@ function bytesToUuid(buf, offset) {
   const bth = byteToHex;
 
   // join used to fix memory issue caused by concatenation: https://bugs.chromium.org/p/v8/issues/detail?id=3175#c4
```
ctavan, line 16: Did you double-verify that the above mentioned issue is no longer present? Either remove the comment or revert to the previous join-solution.

ctavan:
I think we are starting to change too much stuff in this PR, this no longer feels controllable.

Could you please revert your latest code optimizations (let's discuss them in a separate pull request)?

I would then try to get this merged quickly so that we can go through further code optimizations in a more step-by-step manner and therefore controlling better for possible performance optimizations.

So for the time being I'd prefer if this PR would contain only:

* code style changes (eslint config)
* let/const changes
* typed array changes

Let's move all other code changes into a separate pull request.

ctavan:
Oh, and to get this in it would be great if you could [Allow changes to a pull request branch created from a fork](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork). This would help me get this merged and keep all references to your contributions.

ctavan:
Benchmark master:

```
uuidv1() x 1,377,681 ops/sec ±1.47% (91 runs sampled)
uuidv1() fill existing array x 4,881,829 ops/sec ±2.99% (89 runs sampled)
uuidv4() x 306,412 ops/sec ±2.56% (85 runs sampled)
uuidv4() fill existing array x 362,883 ops/sec ±3.90% (85 runs sampled)
uuidv3() x 111,873 ops/sec ±2.63% (81 runs sampled)
uuidv5() x 110,720 ops/sec ±2.73% (82 runs sampled)
Fastest is uuidv1() fill existing array
```

Benchmark this branch:

```
uuidv1() x 1,555,686 ops/sec ±2.63% (90 runs sampled)
uuidv1() fill existing array x 7,804,665 ops/sec ±1.92% (87 runs sampled)
uuidv4() x 311,104 ops/sec ±2.22% (82 runs sampled)
uuidv4() fill existing array x 364,342 ops/sec ±4.73% (86 runs sampled)
uuidv3() x 117,100 ops/sec ±3.78% (82 runs sampled)
uuidv5() x 115,292 ops/sec ±1.76% (83 runs sampled)
Fastest is uuidv1() fill existing array
```

ctavan:
@awwit I just realized that I already hat push access to your branch. So I did the requested cleanups and merged your changes 🎉 

I copied the advanced optimizations to https://github.com/uuidjs/uuid/tree/awwit-refactorings feel free to pick them up from there and create another pull request if you want.

I'm happy to merge those changes as well once I understand their benefit and performance implications.

awwit:
@ctavan Fine! Then I will continue to do in another brunch. I will remove the code that greatly inflates the bundle. And I’ll do something else.

PR #435, opened 2020-05-05
feat: improved v4 performance

awwit:
Performance is increased by using a single `rnds8` buffer to generate random numbers (as in browser version).

Changes affect only the `v1` and `v4`.

Benchmark master:

```
uuidv1() x 1,640,551 ops/sec ±0.76% (91 runs sampled)
uuidv1() fill existing array x 7,369,911 ops/sec ±0.49% (96 runs sampled)
uuidv4() x 389,271 ops/sec ±1.39% (92 runs sampled)
uuidv4() fill existing array x 407,681 ops/sec ±1.62% (91 runs sampled)
uuidv3() x 142,615 ops/sec ±0.81% (87 runs sampled)
uuidv5() x 141,386 ops/sec ±1.09% (87 runs sampled)
```

Benchmark this branch:

```
uuidv1() x 1,653,982 ops/sec ±0.91% (90 runs sampled)
uuidv1() fill existing array x 7,383,926 ops/sec ±0.66% (93 runs sampled)
uuidv4() x 471,989 ops/sec ±0.53% (93 runs sampled)
uuidv4() fill existing array x 484,766 ops/sec ±0.48% (96 runs sampled)
uuidv3() x 141,146 ops/sec ±1.18% (85 runs sampled)
uuidv5() x 147,054 ops/sec ±1.13% (90 runs sampled)
```

### DIFF
--- a/src/v4.js
+++ b/src/v4.js
@@ -2,10 +2,10 @@ import rng from './rng.js';
 import bytesToUuid from './bytesToUuid.js';
 
 function v4(options, buf, offset) {
-  const i = (buf && offset) || 0;
+  const start = (buf && offset) || 0;
 
   if (typeof options === 'string') {
-    buf = options === 'binary' ? new Uint32Array(16) : null;
+    buf = options === 'binary' ? new Uint8Array(16) : null;
     options = null;
   }
 
@@ -19,12 +19,14 @@ function v4(options, buf, offset) {
 
   // Copy bytes to buffer, if provided
   if (buf) {
-    for (let ii = 0; ii < 16; ++ii) {
-      buf[i + ii] = rnds[ii];
+    for (let i = 0; i < 16; ++i) {
+      buf[start + i] = rnds[i];
     }
+
+    return buf;
   }
 
-  return buf || bytesToUuid(rnds);
+  return bytesToUuid(rnds);
 }
 
 export default v4;
--- a/src/v4.js
+++ b/src/v4.js
@@ -2,8 +2,6 @@ import rng from './rng.js';
 import bytesToUuid from './bytesToUuid.js';
 
 function v4(options, buf, offset) {
-  const start = (buf && offset) || 0;
-
   if (typeof options === 'string') {
     buf = options === 'binary' ? new Uint8Array(16) : null;
     options = null;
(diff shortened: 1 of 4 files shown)


src/v4.js
```diff
@@ -19,12 +19,14 @@ function v4(options, buf, offset) {
...
-      buf[i + ii] = rnds[ii];
+    for (let i = 0; i < 16; ++i) {
+      buf[start + i] = rnds[i];
     }
+
+    return buf;
   }
 
-  return buf || bytesToUuid(rnds);
+  return bytesToUuid(rnds);
```
broofa, line 29: Gotta return `buf` if it's passed in.

```suggestion
  return buf || bytesToUuid(rnds);
```

awwit: @broofa sure, I can add test for this case.

But the code can be seen that the buffer is returned after it is filled:
https://github.com/uuidjs/uuid/pull/435/files#diff-8988f35cbc1e6b5ea7f9cb52a4bfa0f2R26

awwit: @broofa there is already tests for this case:

https://github.com/uuidjs/uuid/blob/master/test/unit/v4.test.js#L46-L73

src/v4.js
```diff
@@ -2,10 +2,10 @@ import rng from './rng.js';
 import bytesToUuid from './bytesToUuid.js';
 
 function v4(options, buf, offset) {
-  const i = (buf && offset) || 0;
+  const start = (buf && offset) || 0;
```
ctavan, line 5: Define this variable only in the `if(buf)` scope below, it's not needed outside. It then simplifies to `offset | 0`.

An alternative would be to use default arguments. We'd have to check how efficient the bable'd code is then though (in terms of bundlesize).

src/v4.js
```diff
@@ -2,10 +2,10 @@ import rng from './rng.js';
...
 
   if (typeof options === 'string') {
```
ctavan, line 7: It would be nice to finally get rid of this legacy old API:

```js
uuidv4('binary');
```

Out of scope for this PR, but I've taken note in #437.

src/v4.js
ctavan: @awwit I just realized that we currently do two things: fill the passed buffer and return it. But we only test for filling, not for the returned buffer.

Since we don't know if anybody relies on the fact, that our methods fill the provided buffer and return it, it would be a good opportunity to add another test case that test for the return value when a buffer is passed.

src/v4.js
awwit: @ctavan ok, I will do it a bit later (deprecated)

PR #438, opened 2020-05-06
test: add tests for buffer api

ctavan:
* Simplify the v3/v5 tests for the buffer api
* Add test cases to v1/v3/v5 to ensure that if passing a buffer as
  second parameter that buffer is not only filled but also returned. For
  v4 this has already been added in #435.


ctavan:
@awwit feel free to have a look at this one as well.

PR #439, opened 2020-05-06
feat: improved v35 performance

awwit:
For `v35`, I did a faster parsing of uuid value. This greatly affected the performance.
When parsing a value, it is also validated.

But the bundle size is slightly exceeded…

t’s up to you to decide if this performance increase (and uuid validation) is worth increasing build size.

(Your current solution does not validate uuid strings)

Benchmark master:

```
uuidv1() x 1,644,613 ops/sec ±1.21% (91 runs sampled)
uuidv1() fill existing array x 7,209,045 ops/sec ±0.82% (91 runs sampled)
uuidv4() x 432,989 ops/sec ±0.91% (92 runs sampled)
uuidv4() fill existing array x 454,731 ops/sec ±0.74% (93 runs sampled)
uuidv3() x 137,231 ops/sec ±0.78% (86 runs sampled)
uuidv5() x 135,756 ops/sec ±1.27% (85 runs sampled)
```

Benchmark this branch:

```
uuidv1() x 1,645,002 ops/sec ±1.09% (89 runs sampled)
uuidv1() fill existing array x 7,294,826 ops/sec ±0.82% (92 runs sampled)
uuidv4() x 457,979 ops/sec ±0.76% (90 runs sampled)
uuidv4() fill existing array x 466,131 ops/sec ±0.76% (91 runs sampled)
uuidv3() x 253,029 ops/sec ±1.01% (83 runs sampled)
uuidv5() x 254,426 ops/sec ±1.22% (84 runs sampled)
```

### DIFF
--- a/src/v35.js
+++ b/src/v35.js
@@ -6,11 +6,11 @@ function hexSymToDecNum(n) {
     return n - 48;
   }
   // ----------- A -------- F
-  else if (n >= 65 && n <= 70) {
+  if (n >= 65 && n <= 70) {
     return n - 55;
   }
   // ----------- a -------- f
-  else if (n >= 97 && n <= 102) {
+  if (n >= 97 && n <= 102) {
     return n - 87;
   }
 
@@ -23,30 +23,32 @@ function hexSymToDecNum(n) {
 function uuidToBytes(uuid) {
   const bytes = [];
 
-  if (uuid.length === 36) {
-    for (let i = 0; i < uuid.length; ++i) {
-      let h = uuid.charCodeAt(i);
+  if (uuid.length !== 36) {
+    return bytes;
+  }
 
-      if (i === 8 || i === 13 || i === 18 || i === 23) {
-        // ----- '-'
-        if (h === 45) {
-          continue;
-        } else {
-          return [];
-        }
-      }
+  for (let i = 0; i < 36; ++i) {
+    let h = uuid.charCodeAt(i);
 
-      const l = hexSymToDecNum(uuid.charCodeAt(i + 1));
-      h = hexSymToDecNum(h);
-
-      if (l === -1 || h === -1) {
+    if (i === 8 || i === 13 || i === 18 || i === 23) {
+      // ----- '-'
+      if (h === 45) {
+        continue;
+      } else {
         return [];
       }
-
-      bytes.push(h * 16 + l);
-
-      ++i;
     }
+
+    const l = hexSymToDecNum(uuid.charCodeAt(i + 1));
+    h = hexSymToDecNum(h);
+
+    if (l === -1 || h === -1) {
+      return [];
+    }
+
+    bytes.push(h * 16 + l);
+
+    ++i;
   }
 
   return bytes;
--- a/src/v35.js
+++ b/src/v35.js
@@ -1,12 +1,53 @@
 import bytesToUuid from './bytesToUuid.js';
 
+function hexSymToDecNum(n) {
+  // ------ 0 -------- 9
+  if (n >= 48 && n <= 57) {
+    return n - 48;
+  }
+  // ----------- A -------- F
+  else if (n >= 65 && n <= 70) {
+    return n - 55;
+  }
+  // ----------- a -------- f
+  else if (n >= 97 && n <= 102) {
+    return n - 87;
+  }
+
+  return -1;
+}
+
+/**
+ * Validate and parse UUID to bytes array
+ */
 function uuidToBytes(uuid) {
-  // Note: We assume we're being passed a valid uuid string
   const bytes = [];
 
-  uuid.replace(/[a-fA-F0-9]{2}/g, function (hex) {
-    bytes.push(parseInt(hex, 16));
-  });
+  if (uuid.length === 36) {
+    for (let i = 0; i < uuid.length; ++i) {
+      let h = uuid.charCodeAt(i);
+
+      if (i === 8 || i === 13 || i === 18 || i === 23) {
+        // ----- '-'
+        if (h === 45) {
+          continue;
+        } else {
+          return [];
+        }
+      }
+
+      const l = hexSymToDecNum(uuid.charCodeAt(i + 1));
+      h = hexSymToDecNum(h);
+
+      if (l === -1 || h === -1) {
+        return [];
+      }
+
+      bytes.push(h * 16 + l);
+
+      ++i;
+    }
+  }
 
   return bytes;
 }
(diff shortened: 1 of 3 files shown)


awwit:
Bundlewatch result: https://ja2r7.app.goo.gl/R7EhcSrDJemj7yhk6

Please read PR description ^

src/v35.js
```diff
@@ -1,12 +1,53 @@
...
+ */
 function uuidToBytes(uuid) {
-  // Note: We assume we're being passed a valid uuid string
   const bytes = [];
 
-  uuid.replace(/[a-fA-F0-9]{2}/g, function (hex) {
-    bytes.push(parseInt(hex, 16));
-  });
+  if (uuid.length === 36) {
+    for (let i = 0; i < uuid.length; ++i) {
```
ctavan, line 27: Why not `i+=2` instead of `++i` here and another `++i` below?

test/unit/v35.test.js
```diff
@@ -69,7 +69,7 @@ describe('v5', () => {
     assert.equal(v3('hello.example.com', v3.DNS), '9125a8dc-52ee-365b-a5aa-81b0b3681cf6');
     assert.equal(v3('http://example.com/hello', v3.URL), 'c6235813-3ba4-3801-ae84-e0a6ebb7d138');
     assert.equal(
-      v3('hello', '0f5abcd1-c194-47f3-905b-2df7263a084b'),
+      v3('hello', '0f5abcd1-c194-47f3-905b-2df7263a084b'.toUpperCase()),
```
ctavan, line 72: This would be a good opportunity to break up the huge `v3`/`v5` tests into several smaller `test()` cases, Ideally 1 or 2 assertions per test block.

Then I believe that the `toUpperCase()` test would deserve its own test case with a reasonable description like ("accepts uppercase uuid notation for namespace")

ctavan: I also believe we should, just for the sake of consistency, keep the `v3` and `v5` tests in sync and test the same stuff on both variants of the algorithm.

test/unit/v35.test.js
```diff
@@ -120,14 +120,24 @@ describe('v5', () => {
...
+
+    let invalid = false;
+
+    try {
+      v3('hello.example.com', 'zyzwvuts-rqp0-nmlk-jihg-fedcba000000');
+    } catch (err) {
+      invalid = true;
+    }
+
+    assert.ok(invalid, 'v3 namespace should be invalid');
```
ctavan, line 132: [Use `assert.throws()` instead.](https://nodejs.org/api/assert.html#assert_assert_throws_fn_error_message)

test/unit/v35.test.js
```diff
@@ -69,7 +69,7 @@ describe('v5', () => {
     assert.equal(v3('hello.example.com', v3.DNS), '9125a8dc-52ee-365b-a5aa-81b0b3681cf6');
     assert.equal(v3('http://example.com/hello', v3.URL), 'c6235813-3ba4-3801-ae84-e0a6ebb7d138');
     assert.equal(
```
ctavan, line 71: Move to `assert.strictEqual()`

src/v35.js
```diff
@@ -1,12 +1,53 @@
...
+ * Validate and parse UUID to bytes array
+ */
 function uuidToBytes(uuid) {
-  // Note: We assume we're being passed a valid uuid string
   const bytes = [];
 
-  uuid.replace(/[a-fA-F0-9]{2}/g, function (hex) {
-    bytes.push(parseInt(hex, 16));
-  });
+  if (uuid.length === 36) {
```
ctavan, line 26: Prefer early return to save one level of indentation:

```javascript
if (uuid.length !== 36) {
  return bytes;
}
```

src/v35.js
awwit: @ctavan because not in all cases it is necessary to increase by 2. If we find a `-`, then increase only by 1.
https://github.com/uuidjs/uuid/pull/439/files#diff-4c3fa2fbfa9d4b899d8be4b79ae6907fR33

ctavan: Right…

src/v35.js
```diff
@@ -1,12 +1,53 @@
 import bytesToUuid from './bytesToUuid.js';
 
+function hexSymToDecNum(n) {
```
ctavan, line 3: Is this so much faster than `parseInt(n, 16)`?

awwit: Yes, faster than `parseInt(uuid.substr(i, 2), 16)`
You can check it yourself.

ctavan: This is ridiculous 😂 🤷‍♂️

awwit: Noticeably faster, otherwise I would have done through `parseInt` %)

ctavan: Don't get me wrong: I didn't mean to question, that your solution is considerably faster, I was just amazed by the fact…

awwit:
I completed the validation using the sample from [UUID class](https://github.com/uuidjs/uuid/blob/es6_modules/UUID.js).

This of course increased the size of the module. But not more than 1 Kb (for complete bundle). Why do you have such strict limits? )

It’s a little strange that you care so much about the size of your module’s bundle. Since it includes a little-used `v35`.

It would be logical to publish versions separately (v1, v4, v35): `@uuid/v1`, `@uuid/v4`, `@uuid/v35`.

Of course you understand that the new functionality requires more code (and larger size of bundle).

broofa:
Sorry, I think there may have been a misunderstanding.  I wasn't suggesting that we add the UUID class' code to this PR. Rather, I was simply using that as an example of where we might leverage your code _if/when we decided to add support for a parse/validate feature_.  Historically, however, we've shied away that (too small an audience, already addressed in other modules).

awwit:
@broofa now it’s clear ... then I will try to make the fastest parsing of the UUID that will fit into the limits you set.

ctavan:
@awwit I believe there might still be a misinterpretation: with regards to UUID parsing we‘re really not concerned about bundle size! We‘re more concerned about the feature and API scope of this library! This library is used by millions of projects in the node.js ecosystem to generate UUIDs. If you look at how widespread UUID parser libraries on npm are you will notice that this is an almost negligible fraction compared to the usage of this library.

Hence, we are reluctant of adding little-used features in here that would probably be better off in a separate, special purpose npm module. I think we would be happy to host such an effort in the uuidjs GitHub organization, but for the time being not in this module itself.

And regarding bundlesize: since we moved to ES modules this library supports treeshaking. So people will only get those algorithms in their frontend bundle that they actually use. For Node.js bundle size is not a concern, hence no need to publish separate npm modules.

awwit:
@ctavan good =) then which option will suit you?

In this PR, initially I did only a quick parsing of the UUID and checking for string validation by template `00000000-0000-0000-0000-000000000000`. To match the exception ["namespace must be uuid string or an Array of 16 byte values"](https://github.com/uuidjs/uuid/pull/439/files#diff-4c3fa2fbfa9d4b899d8be4b79ae6907fR117)

It seemed correct to me to correspond to the description of the exception. =)

broofa:
Closing, with apologies in advance if that comes across as rude.  That is not my intent.  To recap the reasons for this:

* The increased code complexity is not warranted.  It doesn't offer enough benefit to the existing `uuid` audience (mainly because this only affects ~1% of users that use v3/v5 uuids).
* High performance parsing will be more interesting if/when it is exposed as part of a module's public API.
* Currently we (uuidjs org) don't have plans for that.  If that changes, we should consider using this implementation.

Thanks!

PR #449, opened 2020-05-12
fix: export package.json required by react-native and bundlers

### LINKED ISSUE #444: package.json is not defined by "exports"
## Describe the bug

'./package.json' is not defined by "exports" when upgrading from v7.0.3 to v8.0.0

## To Reproduce

1. upgrading from v7.0.3 to v8.0.0 and running npm build in a Svelte project
2. See error

## Expected behavior

A rollup bundle without an error as shown below.

## Runtime

- OS: macOS
- Runtime: Node.js
- Runtime Version: 14.1.0

## Additional context

This is the command line output after upgrading from v7.0.3 to v8.0.0 in a Svelte project

```
> Package subpath './package.json' is not defined by "exports" in /Users/sa/0/edm/node_modules/uuid/package.json
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /Users/sa/0/edm/node_modules/uuid/package.json
    at applyExports (internal/modules/cjs/loader.js:491:9)
    at resolveExports (internal/modules/cjs/loader.js:507:23)
    at Function.Module._findPath (internal/modules/cjs/loader.js:635:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:1007:27)
    at Function.requireRelative.resolve (/Users/sa/0/edm/node_modules/require-relative/index.js:30:17)
    at tryResolve (/Users/sa/0/edm/node_modules/rollup-plugin-svelte/index.js:50:19)
    at Object.resolveId (/Users/sa/0/edm/node_modules/rollup-plugin-svelte/index.js:177:21)
    at /Users/sa/0/edm/node_modules/rollup/dist/shared/rollup.js:18217:25
```

I found some additional information in a SO post that might very well be related to the issue I'm seeing https://stackoverflow.com/questions/61621828/getting-strange-error-using-uuid-npm-module-what-am-i-doing-wrong/61730597#61730597

ctavan:
It appears that react-native and some bundlers rely on being able to
introspect the package.json file of a npm module. After adding the
`exports` field to package.json, only the paths defined in that field
are accessible to Node.js. In order to support introspection of the
package.json file itself it has to be listed as an export as well (see:
https://github.com/ai/nanoevents/issues/44#issuecomment-602010343).

It is currently being discussed to change Node.js' behavior again, see:
https://github.com/nodejs/node/issues/33460

Fixes #444

### DIFF
--- a/examples/node-commonjs/example.js
+++ b/examples/node-commonjs/example.js
@@ -45,3 +45,7 @@ console.log('uuid.v3() MY_NAMESPACE', uuid.v3('Hello, World!', MY_NAMESPACE));
 console.log('uuid.v5() DNS', uuid.v5('hello.example.com', uuid.v5.DNS));
 console.log('uuid.v5() URL', uuid.v5('http://example.com/hello', uuid.v5.URL));
 console.log('uuid.v5() MY_NAMESPACE', uuid.v5('Hello, World!', MY_NAMESPACE));
+
+// Some tools like react-native need to introspect the package.json file
+const pkg = require('uuid/package.json');
+console.log('pkg.name', pkg.name);
--- a/examples/node-esmodules/example.mjs
+++ b/examples/node-esmodules/example.mjs
@@ -1,5 +1,6 @@
 import { v1 as uuidv1, v4 as uuidv4, v3 as uuidv3, v5 as uuidv5 } from 'uuid';
 import * as uuid from 'uuid';
+import pkg from 'uuid/package.json';
 
 console.log('uuidv1()', uuidv1());
 
@@ -41,3 +42,6 @@ console.log('uuid.v3() MY_NAMESPACE', uuid.v3('Hello, World!', MY_NAMESPACE));
 console.log('uuid.v5() DNS', uuid.v5('hello.example.com', uuid.v5.DNS));
 console.log('uuid.v5() URL', uuid.v5('http://example.com/hello', uuid.v5.URL));
 console.log('uuid.v5() MY_NAMESPACE', uuid.v5('Hello, World!', MY_NAMESPACE));
+
+// Some tools like react-native need to introspect the package.json file
+console.log('pkg.name', pkg.name);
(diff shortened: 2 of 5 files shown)


ctavan:
> Something about this smells a bit off. I don't really consider `package.json` to be part of our public API. Thus, adding it to `exports` seems wrong. But I guess there's no alternative. 'Really wish RN had figured out a different approach that didn't require 100's or 1000's of projects to all make the same change. :-(

I think this is only half-true: We have been using the `browser` field of `package.json` for module bundlers for ages, I believe the spec for it (by @defunctzombie) even emerged out of this very use case that we had to provide different rng and hashing functions for the browser! So in reality parts of the `package.json` have been part of the public API of this package for a long time.

While I also believe that RN should degrade gracefully if it cannot read the `package.json` (apparently webpack and rollup didn't have any problems) I still think that the fact that now it's no longer only webpack and rollup but also RN who need to `package.json` to work well with this library is not a big game changer.

Speaking of: @TrySound do you have any clue why `rollup` and `webpack` worked just fine (continuing to respect the `browser` field from `package.json`) but RN requires this change?

TrySound:
rollup and webpack use only fs calls for resolving which is not harmed by exports isolation. Metro may use require or require.resolve for some reason but I'm not confident about as never used before.

ctavan:
Given the discussion in https://github.com/nodejs/node/issues/33460 I believe the correct thing to do right now is to treat our `package.json` as part of the public API, at least this section should be considered as such since it defines how module bundlers and Node.js environments should import this module: https://github.com/uuidjs/uuid/blob/a21e4d8c539bd7bfad29e360ab69766f941667b6/package.json#L19-L31

If node changes behavior of `package.json` exports in the future we can still adjust. For the time being exporting `package.json` should lead to minimum breakage in other environments.

abohosam1772:
[]()

PR #453, opened 2020-05-21
feat: slightly improved v1 perf

awwit:
Hello!

I managed to increase the `uuidv1` generation performance by replacing `[]` to `new Array(16)`.

I also added an error code when throwing an exception for exceeding `v1` generation limit per second.
I would like to add error codes to all your exceptions. To make it easier to identify the error by its code (instead of a message).
But we need to work on the names of the error codes. We need to standardize names (or somewhere to look).

If at all you like the idea with error codes =)
If not, then I can remove them.

Benchmark master:

```
uuidv1() x 1,634,946 ops/sec ±0.82% (90 runs sampled)
uuidv1() fill existing array x 7,727,514 ops/sec ±0.63% (92 runs sampled)
uuidv4() x 449,289 ops/sec ±0.63% (95 runs sampled)
uuidv4() fill existing array x 463,831 ops/sec ±0.70% (92 runs sampled)
uuidv3() x 144,401 ops/sec ±0.86% (87 runs sampled)
uuidv5() x 147,395 ops/sec ±1.17% (88 runs sampled)
```

Benchmark this branch:

```
uuidv1() x 1,723,775 ops/sec ±1.21% (92 runs sampled)
uuidv1() fill existing array x 7,763,768 ops/sec ±0.98% (91 runs sampled)
uuidv4() x 425,037 ops/sec ±0.49% (93 runs sampled)
uuidv4() fill existing array x 444,126 ops/sec ±0.52% (95 runs sampled)
uuidv3() x 142,065 ops/sec ±0.85% (88 runs sampled)
uuidv5() x 142,814 ops/sec ±1.04% (88 runs sampled)
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -13,6 +13,12 @@
   },
   "parser": "babel-eslint",
   "rules": {
-    "no-var": ["error"]
+    "no-var": ["error"],
+    "no-empty": [
+      "error",
+      {
+        "allowEmptyCatch": true
+      }
+    ]
   }
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -13,12 +13,6 @@
   },
   "parser": "babel-eslint",
   "rules": {
-    "no-var": ["error"],
-    "no-empty": [
-      "error",
-      {
-        "allowEmptyCatch": true
-      }
-    ]
+    "no-var": ["error"]
   }
 }
--- a/examples/benchmark/benchmark.js
+++ b/examples/benchmark/benchmark.js
@@ -23,11 +23,7 @@ suite
   .add('uuidv1() fill existing array', function () {
     try {
       uuidv1(null, array, 0);
-    } catch (err) {
-      if (err.code !== 'UUID_V1_LIMIT_PER_SEC') {
-        throw err;
-      }
-    }
+    } catch (err) {}
   })
   .add('uuidv4()', function () {
     uuidv4();
--- a/examples/benchmark/benchmark.js
+++ b/examples/benchmark/benchmark.js
@@ -21,7 +21,13 @@ suite
     uuidv1();
   })
   .add('uuidv1() fill existing array', function () {
-    uuidv1(null, array, 0);
+    try {
+      uuidv1(null, array, 0);
+    } catch (err) {
+      if (err.code !== 'UUID_V1_LIMIT_PER_SEC') {
+        throw err;
+      }
+    }
   })
   .add('uuidv4()', function () {
     uuidv4();
--- a/examples/benchmark/benchmark.js
+++ b/examples/benchmark/benchmark.js
@@ -23,7 +23,11 @@ suite
   .add('uuidv1() fill existing array', function () {
     try {
       uuidv1(null, array, 0);
-    } catch (err) {}
+    } catch (err) {
+      // The spec (https://tools.ietf.org/html/rfc4122#section-4.2.1.2) defines that only 10M/s v1
+      // UUIDs can be generated on a single node. This library throws an error if we hit that limit
+      // (which can happen on modern hardware and modern Node.js versions).
+    }
   })
   .add('uuidv4()', function () {
     uuidv4();
--- a/src/bytesToUuid.js
+++ b/src/bytesToUuid.js
@@ -2,40 +2,36 @@
  * Convert array of 16 byte values to UUID string format of the form:
  * XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
  */
-const byteToHex = [];
+const bth = [];
 
 for (let i = 0; i < 256; ++i) {
-  byteToHex.push((i + 0x100).toString(16).substr(1));
+  bth.push((i + 0x100).toString(16).substr(1));
 }
 
-function bytesToUuid(buf, offset) {
-  const i = offset || 0;
-
-  const bth = byteToHex;
-
+function bytesToUuid(buf, offset = 0) {
   // Note: Be careful editing this code!  It's been tuned for performance
   // and works in ways you may not expect. See https://github.com/uuidjs/uuid/pull/434
   return (
-    bth[buf[i + 0]] +
-    bth[buf[i + 1]] +
-    bth[buf[i + 2]] +
-    bth[buf[i + 3]] +
+    bth[buf[offset + 0]] +
+    bth[buf[offset + 1]] +
+    bth[buf[offset + 2]] +
+    bth[buf[offset + 3]] +
     '-' +
-    bth[buf[i + 4]] +
-    bth[buf[i + 5]] +
+    bth[buf[offset + 4]] +
+    bth[buf[offset + 5]] +
     '-' +
-    bth[buf[i + 6]] +
-    bth[buf[i + 7]] +
+    bth[buf[offset + 6]] +
+    bth[buf[offset + 7]] +
     '-' +
-    bth[buf[i + 8]] +
-    bth[buf[i + 9]] +
+    bth[buf[offset + 8]] +
+    bth[buf[offset + 9]] +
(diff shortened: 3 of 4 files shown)


src/bytesToUuid.js
```diff
@@ -2,40 +2,36 @@
  * Convert array of 16 byte values to UUID string format of the form:
  * XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
  */
-const byteToHex = [];
+const bth = [];
```
ctavan, line 5: I preferred the more self-explaining name `byteToHex`

src/v1.js
```diff
@@ -16,7 +16,7 @@ let _lastNSecs = 0;
 // See https://github.com/uuidjs/uuid for API details
 function v1(options, buf, offset) {
   let i = (buf && offset) || 0;
-  const b = buf || [];
+  const b = buf || new Array(16);
```
ctavan, line 19: In `v4` we now use a `Uint8Array(16)`, why not here?
https://github.com/uuidjs/uuid/blob/a361e29a68bbad78318715e6aee696f0006f4b34/src/v4.js#L6

examples/benchmark/benchmark.js
```diff
@@ -21,7 +21,13 @@ suite
...
   })
   .add('uuidv1() fill existing array', function () {
-    uuidv1(null, array, 0);
+    try {
+      uuidv1(null, array, 0);
+    } catch (err) {
+      if (err.code !== 'UUID_V1_LIMIT_PER_SEC') {
+        throw err;
+      }
+    }
```
broofa, line 30: Revert?

Why is this needed?  Are we hitting the rate limit with our benchmark?  If so, that's kind of cool... but it also invalidates the benchmark so should probably be `abort()`ed with an appropriate message.

Note: Silently dropping errors in `catch` blocks is a very strong code smell.  At a minimum, there should be a comment explaining the rationale for it.

src/v1.js
```diff
@@ -70,7 +72,9 @@ function v1(options, buf, offset) {
 
   // Per 4.2.1.2 Throw error if too many uuids are requested
   if (nsecs >= 10000) {
-    throw new Error("uuid.v1(): Can't create more than 10M uuids/sec");
+    const error = new Error("uuid.v1(): Can't create more than 10M uuids/sec");
+    error.code = 'UUID_V1_LIMIT_PER_SEC';
+    throw error;
```
broofa, line 77: Revert.

`code`  is a non-standard error property, and so better to avoid.  Yes, other modules do this but typically it's for the purpose of forwarding codes in some underlying API.

If this is needed (which I'm on the fence about, and 'would like to get @ctavan's opinion here) I'd suggest something along these lines...

----
```javascript
// Create minimal Error subclass (Define in UuidRateError.js file so it can be exported as part of src/index.js?)
export class UuidRateError extends Error {}

// ... Throw in this file when rate exceeded ...
throw new UuidRateError(`UUID creation rate limit exceeded`)

// ... use `instanceof` elsewhere to detect... 
import {v1 as uuidv1, UuidRateError} from 'uuid'
// ...
try {
  uuidv1()
} catch (err) {
  if (err instanceof UuidRateError) // etc...
}
```

src/v1.js
awwit: @ctavan typed array does not increase performance (in this case) =(
Performance remains at the level with an empty array `[]`.

src/v1.js
awwit: @broofa it will be harder for library users. They will need to guess that non-standard exceptions are thrown. When it is already accepted in nodejs to identify an error by code.

https://nodejs.org/api/errors.html#errors_error_code

examples/benchmark/benchmark.js
awwit: @broofa periodically the limit is exceeded on my machine and an exception is thrown.
This is strange, but the block `try/catch` does not affect the performance %)
If an exception is thrown, the following tests will fail.

In new versions of nodejs (v13.14.0), performance is even higher. Exceptions are thrown more often.

broofa: > periodically the limit is exceeded on my machine

This is surprising.  It's the first time I've seen this limit hit in the real world.  (I added that rate check years ag to comply with the spec, never really expecting to hit it in practice.)  What hardware are you running on if I may ask?

src/v1.js
broofa: Huh... I wasn't aware `code` was used by node core.  Interesting.

src/v1.js
ctavan: That performance implication is really interesting/unexpected. [Stackoverflow says](https://stackoverflow.com/a/24855023/1053532):

> Modern engines will use true arrays behind the scenes even if you use Array if they think they can, falling back on property map "arrays" if you do something that makes them think they can't use a true array.

However I just realized that this line in `v4` was deprecated (and never used) anyways, so please ignore my initial comment. See #454

Let's simply leave `new Array(16)` in place. Pre-allocating the size of the array definitely is a reasonable optimization.

src/v1.js
```diff
@@ -70,7 +72,9 @@ function v1(options, buf, offset) {
 
   // Per 4.2.1.2 Throw error if too many uuids are requested
   if (nsecs >= 10000) {
-    throw new Error("uuid.v1(): Can't create more than 10M uuids/sec");
+    const error = new Error("uuid.v1(): Can't create more than 10M uuids/sec");
+    error.code = 'UUID_V1_LIMIT_PER_SEC';
```
ctavan, line 76: If we keep this, I suggest to align it with [Node.js error code naming](https://nodejs.org/api/errors.html#nodejs-error-codes) and call ist something like `ERR_UUID_V1_RATE_EXCEEDED`.

Since, as explained below, this is the only runtime error of this library, I would prefer to revert this for now (or extract this into a separate PR so we can merge the perf optimization first and think about error codes a bit more).

examples/benchmark/benchmark.js
awwit: @broofa I have my own computer on which I run tests (nodejs v12.16.3):

CPU: AMD Ryzen 7 1700
RAM: 32 Gb 2999 MHz

Typically, an exception occurs the first few times the benchmark is launched. And then the testing goes fine. Maybe it's a bug. (With `try/catch` I never get a performance close or higher to 10M)

The exception is much more likely to occur on new versions of nodejs (v13). Because the overall performance of the library is increasing.

On nodejs v13 I usually get performance: `uuidv1() fill existing array x 8,777,105 ops/sec ±0.82% (96 runs sampled)`.

examples/benchmark/benchmark.js
```diff
@@ -21,7 +21,9 @@ suite
     uuidv1();
   })
   .add('uuidv1() fill existing array', function () {
-    uuidv1(null, array, 0);
+    try {
+      uuidv1(null, array, 0);
+    } catch (err) {}
```
ctavan, line 26: I think in the benchmark it is acceptable to just swallow the error.

However just reading this piece of code without additional knowledge might be really confusing. Please add a comment that explains why you are catching the error:

```js
    // The spec (https://tools.ietf.org/html/rfc4122#section-4.2.1.2) defines that only 10M/s v1
    // UUIDs can be generated on a single node. This library throws an error if we hit that limit
    // (which can happen on modern hardware and modern Node.js versions).
```

.eslintrc.json
```diff
@@ -13,6 +13,12 @@
   },
   "parser": "babel-eslint",
   "rules": {
-    "no-var": ["error"]
+    "no-var": ["error"],
+    "no-empty": [
```
ctavan, line 17: Please don't add a global eslint exception. Instead use an `// eslint-disable-next-line no-empty` comment at the place where you want to ignore this rule!

PR #455, opened 2020-05-25
test: parsing non RFC uuid values

awwit:
Hello!

I added tests for parsing non RFC uuid values. So that next time not to violate the current behavior of the module. (With future `v35` optimizations.)


test/unit/v35.test.js
```diff
@@ -108,6 +108,29 @@ describe('v5', () => {
     assert.ok(v3('hello.example.com', '00000000-0000-0000-0000-000000000000'));
   });
 
+  test('v3 parsing non RFC uuid values', () => {
+    assert.strictEqual(
+      v3(
+        'hello.example.com',
+        // equal '00000000-0000-0000-0000-000000000000'
+        '00000000000000000000000000000000',
```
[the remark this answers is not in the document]
awwit: @ctavan I tried to do it in another PR that you rejected: https://github.com/uuidjs/uuid/pull/439
=)

Now I just want to prepare a module for the next v35 optimization (and not disrupt current behavior).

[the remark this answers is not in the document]
ctavan: Uhm well, we rejected #439 because of [the complexity of the UUID parsing code](https://github.com/uuidjs/uuid/pull/439#issuecomment-627442855).

That doesn't mean that an additional input validation of the UUID v3/v5 namespace parameter would make sense. I'd be happy to accept such a change if it was presented in a non-complex manner!

The thing about the test cases that you are adding here is that they actually reveal _unintended_ behavior, however I believe the test-cases should only document the _intended_ behavior!

So in fact you are just uncovering a flaw in the library and we should rather fix that flaw and turn these test cases into `assert.throws()`.

broofa:
> The thing about the test cases that you are adding here is that they actually reveal unintended behavior

💯 agree.   Codifying this behavior in tests suggests that the ability to accept non-compliant UUIDs is a desirable and intentional trait when, in fact, it's not.   The only reason this works is because I was [lazy](https://github.com/uuidjs/uuid/blob/master/src/v35.js#L4) in my initial implementation, for which I apologize.

Proposal:
* Fix `uuidToBytes` so it only accepts valid UUID strings (have it return `null` otherwise?)
* Invert logic of tests added here to validate they _don't_ pass
* Add a `validate(str)` function to public API (just returns `uuidToBytes(str) !== null`)
* Add a `version(string)` function to public API (returns UUID version number) 
* Put these changes on a "v9" branch (for future version 9 release) since stricter parse behavior may break existing v3/v5 use cases.  But hold off on releasing that branch for a while.   These changes aren't urgent enough to push immediately and I suspect our users may be suffering from a bit of major-release fatigue lately.

I believe this would address most of the concerns we've seen around parsing behavior.

BTW, we might want to consider enforcing the valid-ness of UUIDs using a well-crafted regex that we also export publicly.  There's significant demand and value in such a regex. (Note, however, that such a regex would also need to account for the nil UUID, which most proposed solutions don't).  While it won't have the performance of @awwit 's previous solution, it will have the benefit of being consistent both internally and externally.  For example, it could be used [validate form fields](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) such that any values there could be expected to pass whatever validate() function we provide.

Thoughts?

ctavan:
As a side note, deno exposes such validation methods as well: https://github.com/denoland/deno/blob/master/std/uuid/v5.ts#L13

broofa:
> As a side note, deno exposes such validation methods as well

[Interesting copyright](https://github.com/denoland/deno/blob/master/std/uuid/v1.ts#L1) 🤣

ctavan:
> [Interesting copyright](https://github.com/denoland/deno/blob/master/std/uuid/v1.ts#L1) 🤣

Yeah, they only give credit [in the main module file](https://github.com/denoland/deno/blob/55d36006fc00768d1ddfaff8ea8abe6cd2b19e4b/std/uuid/mod.ts#L1)

ctavan:
@broofa I agree with all of your points. It would be a nice opportunity to finally attack these longstanding ideas around parsing and probably make this module even more "complete".

Exposing the new APIs in a tree-shakeable way should also minimize impact for existing users.

And I agree that we should move these developments into a new branch and let them mature for a while before we release yet another major version.

awwit:
@ctavan @broofa I made the changes that you asked me for.

But I can’t create `v9` branch in your repository to change PR target.

BTW, one more question that I wanted to ask a long time.
Can I change your API to use only `Uint8Array` (`buf` argument) in this places?

https://github.com/uuidjs/uuid/blob/eb6038dda7/src/v1.js#L17
https://github.com/uuidjs/uuid/blob/eb6038dda7/src/v4.js#L4
https://github.com/uuidjs/uuid/blob/eb6038dda7/src/v35.js#L30

awwit:
BTW, new way to parse uuid values shows this performance:

```
uuidv3() x 238,241 ops/sec ±0.88% (92 runs sampled)
uuidv5() x 236,577 ops/sec ±0.71% (95 runs sampled)
```

vs master:

```
uuidv3() x 153,787 ops/sec ±0.53% (91 runs sampled)
uuidv5() x 153,972 ops/sec ±0.40% (90 runs sampled)
```

test/unit/v35.test.js
```diff
@@ -108,6 +108,27 @@ describe('v5', () => {
...
+    assert.throws(() => {
+      v3(
+        'hello.example.com',
+        // equal '00000000-0000-0000-0000-000000000000'
+        '00000000000000000000000000000000',
+      );
+    });
+
+    // During parsing only two hex chars in a row are taken into account
+    // The remaining characters are ignored.
```
ctavan, line 121: This comment is now outdated.

test/unit/v35.test.js
```diff
@@ -108,6 +108,27 @@ describe('v5', () => {
...
+
+    assert.throws(() => {
+      v3(
+        'hello.example.com',
+        // equal '00000000-0000-0000-0000-000000000000'
```
ctavan, line 126: This comment is now outdated.

test/unit/v35.test.js
```diff
@@ -204,6 +225,27 @@ describe('v5', () => {
     assert.ok(v5('hello.example.com', '00000000-0000-0000-0000-000000000000'));
   });
 
+  test('v5 parsing non RFC uuid values', () => {
+    assert.throws(() => {
+      v5(
+        'hello.example.com',
+        // equal '00000000-0000-0000-0000-000000000000'
```
ctavan, line 232: Outdated.

test/unit/v35.test.js
```diff
@@ -204,6 +225,27 @@ describe('v5', () => {
...
+        '00000000000000000000000000000000',
+      );
+    });
+
+    // During parsing only two hex chars in a row are taken into account
+    // The remaining characters are ignored.
```
ctavan, line 238: Outdated.

test/unit/v35.test.js
```diff
@@ -204,6 +225,27 @@ describe('v5', () => {
...
+
+    assert.throws(() => {
+      v5(
+        'hello.example.com',
+        // equal '00000000-0000-0000-0000-000000000000'
```
ctavan, line 243: Outdated.

src/v35.js
```diff
@@ -1,12 +1,29 @@
...
+import validate from './validate.js';
+
+// Int32 to 4 bytes
+function numberToBytes(num, bytes) {
+  const offset = bytes.length;
+  bytes.push(0, 0, 0, 0);
+
+  for (let i = 0; i < 4; ++i) {
+    const byte = num & 0xff;
+    bytes[offset + 3 - i] = byte;
```
ctavan, line 11: I think this line could warrant a comment. Something like fill the 4 appended bytes right-to-left.

ctavan: Or a link to the respective stackoverflow question 😉

src/v35.js
```diff
@@ -1,12 +1,29 @@
...
 
 function uuidToBytes(uuid) {
-  // Note: We assume we're being passed a valid uuid string
   const bytes = [];
 
-  uuid.replace(/[a-fA-F0-9]{2}/g, function (hex) {
-    bytes.push(parseInt(hex, 16));
-  });
+  if (!validate(uuid)) {
+    return bytes;
```
ctavan, line 20: If we pre-allocate the `bytes = Array(16)` then we should `return null` here.

src/v35.js
```diff
@@ -28,8 +45,6 @@ export const URL = '6ba7b811-9dad-11d1-80b4-00c04fd430c8';
 
 export default function (name, version, hashfunc) {
   function generateUUID(value, namespace, buf, offset) {
-    const off = (buf && offset) || 0;
-
     if (typeof value === 'string') value = stringToBytes(value);
     if (typeof namespace === 'string') namespace = uuidToBytes(namespace);
```
ctavan, line 49: Since we're at it: I'm wondering why our eslint rules don't strictly demand curly braces? I'd prefer that…

src/version.js
```diff
@@ -0,0 +1,11 @@
+import validate from './validate.js';
+
+function version(uuid) {
+  if (validate(uuid)) {
+    return parseInt(uuid.substr(14, 1), 16);
+  }
+
+  return -1;
```
ctavan, line 8: Here I'm still a bit unsure whether -1 is the best return value for an invalid UUID… @broofa wdyt?

src/v35.js
```diff
@@ -1,12 +1,29 @@
...
+  for (let i = 0; i < 4; ++i) {
+    const byte = num & 0xff;
+    bytes[offset + 3 - i] = byte;
+    num = (num - byte) / 256;
+  }
+}
 
 function uuidToBytes(uuid) {
-  // Note: We assume we're being passed a valid uuid string
   const bytes = [];
```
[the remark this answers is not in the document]
awwit: @ctavan I tested the performance and did not see the difference.
Well, let it be preallocated )

src/v35.js
awwit: @ctavan for optimizations from the JS engine, it's best to always return the same type for each function.
Better to return an empty array…

src/version.js
awwit: @ctavan I would recommend always returning the same type from a function…

.eslintrc.json
```diff
@@ -13,6 +13,7 @@
   },
   "parser": "babel-eslint",
   "rules": {
-    "no-var": ["error"]
+    "no-var": ["error"],
+    "curly": ["error", "all"]
```
broofa, line 17: Why?  Revert?  @ctavan Any opinion here?  Since we're using `prettier`, maybe avoid custom rules that are just about esthetics.

src/v35.js
```diff
@@ -1,12 +1,27 @@
...
+
+// Int32 to 4 bytes https://stackoverflow.com/a/12965194/3684944
+function numberToBytes(num, bytes, offset) {
+  for (let i = 0; i < 4; ++i) {
+    const byte = num & 0xff;
+    // Fill the 4 bytes right-to-left.
+    bytes[offset + 3 - i] = byte;
+    num = (num - byte) / 256;
+  }
+}
```
broofa, line 12: Not needed if we use the `uuidToBytes()` implementation I suggest, below.
```suggestion
```

test/unit/v35.test.js
```diff
@@ -204,6 +219,19 @@ describe('v5', () => {
     assert.ok(v5('hello.example.com', '00000000-0000-0000-0000-000000000000'));
   });
 
+  test('v5 parsing non RFC uuid values', () => {
```
broofa, line 222: same same

src/v35.js
```diff
@@ -1,12 +1,27 @@
...
 
 function uuidToBytes(uuid) {
-  // Note: We assume we're being passed a valid uuid string
-  const bytes = [];
+  if (!validate(uuid)) {
```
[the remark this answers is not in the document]
awwit: @broofa I checked this code performance:

```
uuidv3() x 198,096 ops/sec ±0.86% (94 runs sampled)
uuidv5() x 200,775 ops/sec ±0.55% (95 runs sampled)
```

This code is shorter, but less productive.

src/v35.js
```diff
@@ -1,12 +1,29 @@
...
-    bytes.push(parseInt(hex, 16));
-  });
+  if (!validate(uuid)) {
+    return bytes;
+  }
+
+  numberToBytes(parseInt(uuid.slice(0, 8), 16), bytes);
+  numberToBytes(parseInt(uuid.slice(9, 13) + uuid.slice(14, 18), 16), bytes);
+  numberToBytes(parseInt(uuid.slice(19, 23) + uuid.slice(24, 28), 16), bytes);
+  numberToBytes(parseInt(uuid.slice(28), 16), bytes);
```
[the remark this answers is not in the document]
broofa: FWIW, the perf numbers:

@awwit's implementation:
```
uuidv3() x 215,511 ops/sec ±1.49% (83 runs sampled)
uuidv5() x 217,387 ops/sec ±1.24% (81 runs sampled)
```
My suggested implementation:
```
uuidv3() x 177,188 ops/sec ±3.32% (89 runs sampled)
uuidv5() x 184,682 ops/sec ±1.44% (83 runs sampled)
```

Classic tradeoff between conciseness vs. performance, I guess?  @ctavan thoughts?

src/v35.js
broofa: Whups, just seeing this.    As I said in my other comment, I'd err on the side of using default `prettier` rules, and avoiding custom rules simply for aesthetics.  Not something I feel strongly about though.

src/regex.js
```diff
@@ -0,0 +1,3 @@
+const uuidRegex = /^([0-9a-f]{8}-[0-9a-f]{4}-[1345][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})|((00000000-0000-0000-0000-000000000000))$/i;
```
broofa, line 1:
```suggestion
const REGEX = /^([0-9a-f]{8}-[0-9a-f]{4}-[1345][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})|((00000000-0000-0000-0000-000000000000))$/i;
```

.eslintrc.json
ctavan: re `prettier`: Enforcing curly braces [changes the AST and is therefore not in the scope of `prettier`](https://github.com/prettier/prettier/issues/4262#issuecomment-379000483).

Re this setting: I was actually hoping that the `"standard"` eslint config would give us a reasonable set of default rules like this one but apparently that's not the case. I personally like https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb but should we decide to include it we should do that in a different PR, that will go too far. It was probably a bad move from my end to propose adding this change to this PR in the first place, apologies! 🙇‍♂️

src/index.js
```diff
@@ -2,3 +2,6 @@ export { default as v1 } from './v1.js';
 export { default as v3 } from './v3.js';
 export { default as v4 } from './v4.js';
 export { default as v5 } from './v5.js';
+export { default as REGEX } from './regex.js';
```
ctavan, line 5: We should discuss the API a bit more in depth. This would currently imply the following usage:

```js
import { REGEX } from 'uuid';
```

Having a super generic name like `REGEX` in this case is a bit unfortunate. I also find the `vX` method names already a pretty unfortunate choice because they kinda force you to do

```js
```

in order to have a more telling method name further down in the code (who would otherwise guess that `v4()` somewhere further down in the source code generates a UUID?!).

In the [tc39/proposal-uuid](https://github.com/tc39/proposal-uuid#why-call-the-export-randomuuid-and-not-something-like-uuidv4) we therefore proposed the term `randomUUID()` instead of `v4()`.

Suggestion: `UUID_REGEXP` (also `REGEXP` instead of `REGEX` to be closer to how the JavaScript language's `RegExp`)

@awwit @broofa what do you think? (let's discuss this through before changing the code).

src/index.js
```diff
@@ -2,3 +2,6 @@ export { default as v1 } from './v1.js';
...
+export { default as version } from './version.js';
```
ctavan, line 6: Same discussion as for `REGEX` applies here as well… `UUIDVersion()`? Looks a bit weird but would be consistent with `randomUUID()`-style naming…

src/index.js
```diff
@@ -2,3 +2,6 @@ export { default as v1 } from './v1.js';
...
+export { default as validate } from './validate.js';
```
ctavan, line 7: `validateUUID()` would then be an appropriate name maybe?

src/regex.js
```diff
@@ -0,0 +1,3 @@
+const REGEX = /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
+
+export default REGEX;
```
ctavan, line 3: `UUID_REGEXP` see above.

src/v35.js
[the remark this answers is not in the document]
ctavan: I would again favor conciseness over performance in this case.

test/unit/validate.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+    assert.strictEqual(validate('109156be-c4fb-41ea-b1b4-efe1671c5836'), true);
+
+    assert.strictEqual(validate('a981a0c2-68b1-35dc-bcfc-296e52ab01ec'), true);
+
+    assert.strictEqual(validate('90123e1c-7512-523e-bb28-76fab9f2f73d'), true);
+
+    assert.strictEqual(validate('invalid uuid string'), false);
+
+    assert.strictEqual(validate('00000000000000000000000000000000'), false);
```
ctavan, line 18: Edge cases please add `validate('')` and `validate()`

test/unit/validate.test.js
```diff
@@ -0,0 +1,27 @@
+import assert from 'assert';
+import validate from '../../src/validate.js';
+
+describe('validate', () => {
+  test('validate uuid', () => {
+    assert.strictEqual(validate('00000000-0000-0000-0000-000000000000'), true);
```
ctavan, line 6: Great that you extracted these tests as suggested by @broofa !

However, I believe we should keep _one_ test with an invalid namespace in the `v3` and `v5` test suite in order to ensure that the error thrown by `validate()` is in fact propagated and not somehow swallowed by the uuid generation methods.

src/index.js
broofa: FWIW, the `v4` convention came about because prior to object destructing methods were almost always referenced through the module object.  So it was typically `uuid.v4()`.  Including the module name in module properties was unnecessarily redundant.  E.g. `uuid.uuidv4()`.  But... times change.

I suggested `REGEX` because it's in keeping with the pattern we currently promote, where methods are renamed with `import ... as`.  I fully expected we'd document this with something like `import REGEX as UUID_REGEX`.

I'm not wedded to the current names (other than to suggest that we should retain backward compatibility for at least one major release).  If we can agree on a new naming scheme that works well moving forward I'm all in favor of it.  However, we'd be doing all of our users a big favor if the names ended up being consistent with whatever names the standard module proposal ends up going with. (I don't know what those are, however).

Re: REGEX vs REGEXP, the former is [far more common](https://trends.google.com/trends/explore?geo=US&q=regex,regexp).

src/index.js
broofa: [Just deleted a long spiel about how to rename things.]

Out of respect for @awwit, let's not go down the renaming rabbit hole in this PR.  We've already agreed to push this work to `v9` so I think it'd be better to stick with the current custom of terse names for the moment (`version`, `validate`, and `REGEX`), and open a new issue for the renaming discussion.   We could even put something in the README (on `master`) asking people to chime in with their thoughts to help generate discussion.

ctavan: OK, fine for me.

What I would however like to see before we merge is some initial documentation of the new methods in the `README_js.md`. Could you add that @awwit ? I'm happy to merge a first iteration into a new development branch then.

src/index.js
ctavan: OK, let's not change the naming for now and stick with the currently proposed names for the first merge.

test/unit/validate.test.js
awwit: @ctavan It is already there:
https://github.com/uuidjs/uuid/blob/d5ce415363/test/unit/v35.test.js#L99-L109
https://github.com/uuidjs/uuid/blob/d5ce415363/test/unit/v35.test.js#L195-L205

awwit:
@ctavan @broofa I edited the babel configuration a bit. Specified the version of IE for which the module will be built.

https://babeljs.io/docs/en/babel-preset-env

```
esmBrowser: {
  presets: [['@babel/preset-env', { targets: { ie: '11' }, modules: false }]],
},
```

Or is it superfluous?

.babelrc.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
       presets: [['@babel/preset-env', { targets: { node: '8' }, modules: 'commonjs' }]],
     },
     esmBrowser: {
-      presets: [['@babel/preset-env', { modules: false }]],
+      presets: [['@babel/preset-env', { targets: { ie: '11' }, modules: false }]],
```
ctavan, line 9: I just checked again and I think [this is not necessary](https://babeljs.io/docs/en/babel-preset-env#targets): 

> Sidenote, if no targets are specified, @babel/preset-env will transform all ECMAScript 2015+ code by default.

awwit: Is explicit better than implicit? =)

PR #458, opened 2020-05-27
chore: add proseWrap prettier config to wrap markdown

ctavan:
I no longer want to try to achieve consistent line lengths in markdown
so I'd prefer prettier to do that for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,23 +1,34 @@
 # Changelog
 
-All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
+All notable changes to this project will be documented in this file. See
+[standard-version](https://github.com/conventional-changelog/standard-version) for commit
+guidelines.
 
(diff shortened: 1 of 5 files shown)


LinusU:
Personally, I think it's very much better to not wrap lines in `.md` files, and let the editor do that instead.

With this change you have to save the file and run prettier, instead of just having the text flow naturally as you type it in your editor. You are also now forced to 80 chars width instead of the size of your window.

Lastly, I think that the diffs often become much harder to read, when inserting a few words in the beginning of the paragraph changes every line under it. When the paragraph is on one line in the source, GitHub (and others) shows a nice diff that highlights the added words in green, that won't work when multiple lines are changed around...

ctavan:
I think the crucial thing is that this way we enforce a consistent style with the commit hooks or if people don't have them set up properly at least the CI will fail. Otherwise we end up with an inconsistent mess like right now where some paragraphs in the same file apparently have been written with a soft-wrapping editor and others with a hard-wrapping editor…

So in this case I'm with @broofa 😛

LinusU:
Would have preferred to set prettier to put everything on one line then, that way any developer can soft-wrap to whatever width they want... Anyhow, not a big deal...

PR #459, opened 2020-05-28
chore: set prettier proseWrap to never

ctavan:
This basically does the opposite of d5ce415363fcf3f9caab56b2d8a17f341ba51889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,34 +1,23 @@
 # Changelog
 
-All notable changes to this project will be documented in this file. See
-[standard-version](https://github.com/conventional-changelog/standard-version) for commit
-guidelines.
+All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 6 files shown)


ctavan:
Humm, I'm undecided if I find this one better.

For me as a vim-user the problem is that [I can't seem to have proper soft-wrap at a specific column](https://stackoverflow.com/questions/989093/soft-wrap-at-80-characters-in-vim-in-window-of-arbitrary-width) which is super annoying in situation where my vim window happens to be quite wide… 🤷‍♂️ 

Also it looks a bit weird in the case of [LICENSE.md](https://raw.githubusercontent.com/uuidjs/uuid/6d8529a106188ee884450f7a54af7c60918e67d9/LICENSE.md).

WDYT @LinusU @broofa ?

LinusU:
Thank you for taking the time to make this PR after my comments ❤️ 

I personally prefer this very much, I'm sad to see that there were troubles getting it to work with vim. On the other hand, if you previously had a window that was less than 100 characters wide, e.g. 90, it would end up either double wrapping or cutting off the right part of the text...

It's not a hill that I will die on, so if this is an inconvenience for you, then feel free to close it ☺️

broofa:
I'm fine with this.

PR #460, opened 2020-05-29
chore: improve code style

ctavan:
Some changes from #455 that I'd like to see in master.

/cc @awwit


ctavan:
@broofa I would like to merge this subset of #455 into master first before I merge #455 into a new development branch.

Issue #180, opened 2017-02-17
Expose bytes to uuid


wesleytodd:
Do you have any interest in exposing the lib method `bytesToUuid`?

Usecase: I am writing two uuid's to a buffer which I am hex encoding for use in a http header, but when I decode them, I would like to store are traditional string representation in the `XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX` format.

It would be nice to do something like:

```javascript
var buf = Buffer.from(hexEncodedString, 'hex');
var id1 = uuid.bytesToUuid(buf.slice(0, 16));
var id2 = uuid.bytesToUuid(buf.slice(17, 32));
```

Maybe there is another way to do this that I don't know about, so let me know if that is the case.  Otherwise I can submit a PR to expose it.

broofa:
No, this isn't something I expect will ever be part of the public API.

That said, `require('uuid/lib/bytesToUuid')` works.  Just keep in mind that changes that break that will not be considered "breaking changes" from a versioning perspective.

taion:
Would it be possible to reconsider this? With native ES module support in Node, the "deep" import no longer works at all.

It'd be convenient to have a general-purpose way to convert a Node buffer with the bytes for a UUID into the standard hex representation, and using the `bytesToUuid` here seems like one of the nicer ways to do so.

As a comparison, the `UUID` type in the Python standard library exposes a `bytes` parameter when constructing a UUID that allows doing exactly this: https://docs.python.org/3/library/uuid.html#uuid.UUID.

taion:
The absence of an exposed `bytesToUuid` method might tempt one to use `uuidv4({ random: ... })`, which ... can then break subtly when that function is given, say, a UUIDv5, and lead to hard-to-diagnose bugs.

ctavan:
We're currently considering to extend the API surface of this library (https://github.com/uuidjs/uuid/pull/455#issuecomment-634721504), thank you for bringing up that use case.

We'll definitely discuss this!

ctavan:
@wesleytodd @taion we have just released `uuid@8.3.0-beta.0` which includes https://github.com/uuidjs/uuid#uuidstringifyarr-offset would you mind testing it and report if that works for you?

ctavan:
Added in #479

taion:
Passes our test suite. I'll deploy shortly, but I don't expect any problems. Will let you know if anything comes up, but assume it's working fine for us unless you hear otherwise.

[...]

PR #463, opened 2020-06-02
fix: concise uuidToBytes

broofa:
Putting up two possible implementations of `uuidToBytes` for us to consider:

* This PR, `uuidToBytes_concise`, is similar to what's on `master`.  Slightly better perf and compactness, pretty readable.
* PR #464, `uuidToBytes_fast` is @awwit's approach in `v9`, but with the `numberToBytes` stuff done inline.  Yields slightly better performance and saves a few bytes when minified.  Not great readability (as tends to be the case with highly optimized code), hence the extra comments. 

Here are the numbers for the various flavors:

| branch | performance (ops/sec) | size (minified+gzip bytes)<sup>1</sup> |
|---|---|---|
|  `master`<sup>2</sup> | 138-141K | 132 |
| `v9` | 209-221K | 247 |
| `uuidToBytes_concise` | 146-170K | 125 |
| `uuidToBytes_fast` | 228-230K | 183 |

Thoughts?  Pick one, I guess.  :-)

1.  "size" is for just the `uuidToBytes()` implementation run through `jsmin | gzip -c | wc -c`
2. `master` implementation does not currently `validate()` like the other approaches mentioned here do.

### DIFF
--- a/bundlewatch.config.json
+++ b/bundlewatch.config.json
@@ -1,13 +1,13 @@
 {
   "files": [
     { "path": "./examples/browser-rollup/dist/v1-size.js", "maxSize": "0.8 kB" },
-    { "path": "./examples/browser-rollup/dist/v3-size.js", "maxSize": "1.8 kB" },
+    { "path": "./examples/browser-rollup/dist/v3-size.js", "maxSize": "1.9 kB" },
     { "path": "./examples/browser-rollup/dist/v4-size.js", "maxSize": "0.5 kB" },
     { "path": "./examples/browser-rollup/dist/v5-size.js", "maxSize": "1.2 kB" },
 
     { "path": "./examples/browser-webpack/dist/v1-size.js", "maxSize": "1.3 kB" },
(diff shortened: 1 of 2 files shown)


ctavan:
Tough call, I have no strong opinion 🤷‍♂️ 

I'm a huge fan of conciseness but I also think bundlesize is not an issue for v3/5 and performance _is_ a promise that this library gives.

I would make good-enough test coverage the requirement for the complex fast solution (I haven't checked how good the test coverage is for this specific code path).

In the next step we could also consider exposing this method alongside `bytesToUUID` (#180).

wesleytodd:
TBQH, the (code in here](https://github.com/uuidjs/uuid/pull/464/files#diff-4c3fa2fbfa9d4b899d8be4b79ae6907fR11-R37) is not entirely illegible, especially with the comments.  If that is the only concern I would go with it for the small perf and size improvement.

And also, 👍 to exposing it or moving it out to a separate package.  Either one would have solved the issue I had in #180

broofa:
Moving forward with #464

PR #464, opened 2020-06-02
Version 9 improvements

broofa:
This started out as a variation of #463 (uuidToBytes implementation), but now includes the following per discussion in this PR:

* [x] Rename `uuidToBytes()` -> `stringify()`
* [x] Rename `bytesToUuid()` -> `parse()`
* [x] `parse()` (was `uuidToBytes()`) now returns Uint8Array 
* [x] Expose `version()`, `validate()`, `parse()`, and `stringify()` as part of public API 
* [x] Unit tests and benchmarks for `stringify()` and `parse()`
* [ ] Document new methods in README


broofa:
Pushing forward with this approach.  I can move relevant work to #463 if we decide to go with that.

* Pulled uuidToBytes into separate file
* uuidToBytes unit test
* Added uuidToBytes to exports
* Added benchmark for uuidToBytes.

test/unit/uuidToBytes.test.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+  return bytes;
+}
+
+describe('uuidToBytes', () => {
+  test('string->bytes->string symmetry', () => {
+    // Verify string->bytes->string symmetry for a selection of uuids
+    for (let i = 0; i < 1000; i++) {
+      const uuid = uuidv4({ rng });
+      assert.equal(bytesToUuid(uuidToBytes(uuid)), uuid);
```
ctavan, line 21: Do we test `bytesToUuid` separately? If not, then in principle both methods could be buggy in way where the bugs just compensate each other.

To be on the safe side we should add an additional test case where we _explicitly_ test that `uuidToBytes` for a fixed UUID yields a fixed expected byte array.

test/unit/uuidToBytes.test.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+  });
+
+  test('Ignore case', () => {
+    // Verify upper/lower case neutrality
```
ctavan, line 26: Rather move this comment directly into the test description and then remove it.

test/unit/uuidToBytes.test.js
```diff
@@ -0,0 +1,46 @@
...
+function rng(bytes = []) {
+  for (let i = 0; i < 16; i++) {
+    bytes[i] = rand(256);
+  }
+  return bytes;
+}
+
+describe('uuidToBytes', () => {
+  test('string->bytes->string symmetry', () => {
+    // Verify string->bytes->string symmetry for a selection of uuids
```
ctavan, line 18: Rather move this comment directly into the test description and then remove it.

test/unit/uuidToBytes.test.js
```diff
@@ -0,0 +1,46 @@
...
+  test('Ignore case', () => {
+    // Verify upper/lower case neutrality
+    assert.deepStrictEqual(
+      uuidToBytes('0f5abcd1-c194-47f3-905b-2df7263a084b'),
+      uuidToBytes('0f5abcd1-c194-47f3-905b-2df7263a084b'.toUpperCase()),
+    );
+  });
+
+  test('Null UUID', () => {
+    // Verify null Uuid
```
ctavan, line 34: Rather move this comment directly into the test description and then remove it.

src/uuidToBytes.js
```diff
@@ -0,0 +1,39 @@
+import validate from './validate.js';
+
+function uuidToBytes(uuid) {
```
ctavan, line 3: The current UUID proposal at https://github.com/tc39/proposal-uuid#uuid-api always uses all-uppercase `UUID` in method naming. The reasoning was, that acronyms are usually not camel-cased in the JavaScript language, e.g. [`decodeURIComponent()`](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent), [`url.toJSON()`](https://developer.mozilla.org/de/docs/Web/API/URL/toJSON), …

I'd personally prefer to use this kind of naming here as well:

```
UUIDToBytes()
bytesToUUID()
```

If you feel strongly against it I can also live with the current state.

broofa:
Comments look good, @ctavan.  I'm out of town until wednesday, however, so may not be able to get to this until then.

test/unit/uuidToBytes.test.js
broofa: Added explicit byte array test.

src/uuidToBytes.js
broofa: I actually dislike acronyms in method names for exactly this reason.  It's bad enough when the acronym appears in the middle of a name (is it `innerHtml` or `innerHTML`?  I can never remember.  😢), but when it appears at the beginning of function or method name - where you really don't expect uppercase - it's visually jarring.    `UUIDToBytes` just looks weird.  Note, too, that nowhere in Node's built-in API (2500+ method names) is there a method that starts with an uppercase letter.

Maybe the best thing to do here is just emulate the JSON API and rename these to `parse` (uuidToBytes) and `stringify` (bytesToUuid).  Thoughts?

ctavan: Re parse/stringify: Is a byte array the canonical representation of a UUID so that especially `parse` is self-explaining? The only other representation that comes to my mind would be something like `int128` but since something like that isn't part of the JavaScript language anyways I guess it's safe to assume that a byte array is the only sane in-memory representation of a UUID…

Now what about the broader API surface (from the `v9` branch)?

```
UUID.validate()
UUID.version()
UUID.parse()
UUID.stringify()
```

Together with a namespace these all seem to be pretty straightforward and `UUID.bytesToUuid()` would indeed be inconsistent.

It only conflicts a little bit with the current [`randomUUID()` proposal](https://github.com/tc39/proposal-uuid#uuid-api) but since that proposal is in a very early stage I'm not sure if it makes sense at all to have that influence our API evolution within this module.

So yeah, `stringify`/`parse` would be fine for me 👍 .

broofa: > Is a byte array the canonical representation of a UUID so that especially `parse` is self-explaining?

Yes. 

That's what I used as the backing data structure for the `UUID` class [I proposed](https://codepen.io/broofa/pen/KYOvWE?editors=0010) and I found it worked well.  Also, it's what this module used in the [old `parse` and `unparse` API](https://github.com/uuidjs/uuid/tree/v1.4.4#uuidparseid-buffer-offset) it used to export (as `Buffer` instances).

The use of `Buffer` in that old API was mostly because a) this module initially targeted Node and b) typed arrays weren't a thing yet, fwiw.  Clearly that's not the right choice anymore, so I think the only real decision here is should we go with `Array` or `Uint8Array`?  I'm inclined to say the latter.  I.e. We can have `unparse()` accept any array-like[16] input, but `parse()` would only ever return a `Uint8Array`.

ctavan: Sounds reasonable.

ctavan: BTW that's deno's choice as well: https://github.com/denoland/deno/blob/d615f0ff82f5b2b530940a21a61e5b4b08b1d1f4/std/uuid/_common.ts#L40-L47

broofa:
[Adding @LinusU and @TrySound as reviewers, as there's been a bit of feature creep here]

Requesting re-review from @ctavan because switching to Uint8Array for parse() has resulted in a few changes elsewhere in the code.  Also, it's worth noting that this change reduced the benchmark performance of `parse()` by ~25%.

broofa:
@ctavan: Note that the UMD version of the API has `uuid` prefixed on the new methods.  So, `uuidVersion`, `uuidParse`, `uuidParse()`, and `uuidStringify()`.  This is mostly consistent with the existing `uuidv1` methods.  (The unprefixed names felt a bit too generic to export as globals so... 😕 ).  Let me know if you have a better idea for how to handle this.

src/stringify.js
```diff
@@ -0,0 +1,52 @@
...
+
+  // Sanity check for valid UUID.  This works because if any of
+  // the input array values don't map to a defined hex octet, the string length
+  // will get blown out (e.g. "74af23d8-85undefined2c44-...")
+  //
+  // This is a somewhat crude check, but avoids having to check each value
+  // individually.
+  if (uuid.length !== 36) {
+    throw new TypeError(
+      'Invalid result UUID.  Please insure input is array-like, and contains 16 integer values 0-255',
```
ctavan, line 45: _ensure_

test/unit/stringify.test.js
```diff
@@ -0,0 +1,64 @@
...
+  0xf7,
+  0x26,
+  0x3a,
+  0x08,
+  0x4b,
+];
+
+describe('stringify', () => {
+  test('Stringify Array', () => {
+    assert.deepStrictEqual(stringify(BYTES), '0f5abcd1-c194-47f3-905b-2df7263a084b');
```
ctavan, line 25: `strictEqual` should be enough. But I guess `deepStrictEqual` doesn't hurt.

src/v35.js
```diff
@@ -48,19 +23,23 @@ export default function (name, version, hashfunc) {
     }
 
     if (typeof namespace === 'string') {
-      namespace = uuidToBytes(namespace);
+      namespace = parse(namespace);
     }
 
-    if (!Array.isArray(value)) {
-      throw TypeError('value must be an array of bytes');
+    if (!value[Symbol.iterator]) {
```
ctavan, line 29: [Symbol.iterator doesn't seem to be supported in IE11](https://caniuse.com/#search=symbol.iterator).

However babel should take care of that for the browser builds.

ctavan:
> @ctavan: Note that the UMD version of the API has `uuid` prefixed on the new methods. So, `uuidVersion`, `uuidParse`, `uuidParse()`, and `uuidStringify()`. This is mostly consistent with the existing `uuidv1` methods. (The unprefixed names felt a bit too generic to export as globals so... 😕 ). Let me know if you have a better idea for how to handle this.

Not really a better idea. Consistent would be `uuidversion` but that doesn't read well… I consider the umd build really just an optional feature e.g. for educational purposes where people do online coding courses or things like that, so I'm not so much afraid about the api surface on that end.

Also, with cool things like `jspm` there seems to be a modern alternative for the direct browser loading use case.

So maybe just not expose these methods in the umd build at all until somebody complains?

> Edit: Also, what's up with the failing webdriver tests? Is that just browserwatch being flakey?

Test fail in the old browsers:

* IE11
* Firefox 44
* Chrome 49

[`String.padStart()` seems to be the culprit](https://caniuse.com/#feat=pad-start-end):

```
1592943821280:SEVERE:http://127.0.0.1:9000/browser-webpack/dist/all.js 1:991 Uncaught TypeError: o.toString(...).padStart is not a function
,1592943821339:SEVERE:http://127.0.0.1:9000/favicon.ico 0:0 Failed to load resource: the server responded with a status of 404 (Not Found)
,1592943831873:SEVERE:http://127.0.0.1:9000/browser-webpack/dist/v1.js 1:992 Uncaught TypeError: o.toString(...).padStart is not a function
,1592943842443:SEVERE:http://127.0.0.1:9000/browser-webpack/dist/v4.js 1:992 Uncaught TypeError: o.toString(...).padStart is not a function
,1592943852840:SEVERE:http://127.0.0.1:9000/browser-rollup/dist/all.js 1:465 Uncaught TypeError: u.toString(...).padStart is not a function
,1592943863391:SEVERE:http://127.0.0.1:9000/browser-rollup/dist/v1.js 1:465 Uncaught TypeError: u.toString(...).padStart is not a function
,1592943873923:SEVERE:http://127.0.0.1:9000/browser-rollup/dist/v4.js 1:461 Uncaught TypeError: r.toString(...).padStart is not a function
```

src/v35.js
broofa: Removed Symbol.iterator check for the time being.

broofa:
@ctavan Where are you seeing those webdriver error messages?  I'm not seeing anything like that in the github "checks" logs (e.g. https://github.com/uuidjs/uuid/pull/464/checks?check_run_id=801127791).  All I see are timeout errors.

ctavan:
I got those console errors from the Browserstack UI. Last time I think I had issues inviting you to my Open Source team there. Shall I give that another try?

Seems like now it's Chrome 49 and IE 11 that are broken, however this time I don't get meaningful console errors from Browserstack.

However this screenshot: 
image

tells me that something in `v5()` must be broken because all lines up until https://github.com/uuidjs/uuid/blob/b51f172b22580dabd382861c1a6316173244f51b/examples/browser-webpack/example-all.js#L25 seem to render fine.

broofa:
Temporary browserstack workaround: Building and pushing to my personal site, then doing manual testing there.  It's not great but I've figured out the immediate issue (No `Array.from` on IE11).  All checks passing.

src/v35.js
```diff
@@ -48,19 +23,21 @@ export default function (name, version, hashfunc) {
...
+      throw TypeError('Namespace must be array-like (16 iterable integer values, 0-255)');
     }
 
-    if (!Array.isArray(namespace) || namespace.length !== 16) {
-      throw TypeError('namespace must be uuid string or an Array of 16 byte values');
-    }
+    // Compute hash of namespace and value, Per 4.3
+    // Future: Use spread syntax when supported on all platforms, e.g. `bytes =
+    // hashfunc([...namespace, ... value])`
+    let bytes = new Uint8Array(namespace.length + value.length);
```
LinusU, line 36: Hardcoding 16 here could potentially make the minified code a bit smaller...

```suggestion
    let bytes = new Uint8Array(16 + value.length);
```

Issue #451, opened 2020-05-14
ESM Problems with Jest@26 and Node.js 14.x


ctavan:
Hi, I'm using jest to test my project.
Trying to test a file where I import uuid results with an error:
```
 ● Test suite failed to run

    SyntaxError: The requested module 'uuid' does not provide an export named 'v1'

          at async Promise.all (index 0)
      at jasmine2 (node_modules/jest-jasmine2/build/index.js:228:5)
```

I saw @dps910 comment and I guess it's something similar.
I'm using UUID@8, node@14 and Jest@26 (and it's freshly-supported esm modules implementation).

Jest command:
`node --experimental-vm-modules .\node_modules\jest\bin\jest.js`
The file that I'm testing imports uuid like you suggested:
```js
import { v1 as uuidv1 } from 'uuid';
uuidv1();
```

It fails only when I'm testing though.

_Originally posted by @Tzahile in https://github.com/uuidjs/uuid/issues/245#issuecomment-628650062_

TrySound:
Looks like "exports" is not supported yet.
https://github.com/facebook/jest/issues/9430#issuecomment-623941616

ctavan:
@Tzahile as pointed out by @TrySound, ESM support in jest is not complete yet. I think you will have to wait for ESM support in jest to be finished before you can use it with this library.

SimenB:
FWIW, tracking issue: https://github.com/facebook/jest/issues/9771

pl12133:
My workaround for this was to manually map the `uuid` module to it's commonJS export:

```js
// jest.config.js
moduleNameMapper: {
    // Force module uuid to resolve with the CJS entry point, because JEst does not support package.json.exports. See https://github.com/uuidjs/uuid/issues/451
    "uuid": require.resolve('uuid'),
},
```

pl12133:
I can confirm #616 fixes this issue in my project using Jest 28 and jest-environment-jsdom. Thanks for working on it.

grosch:
So that seems like an awesome workaround, but it doesn't work if you're using a jest.config.mjs instead as you can't `require` things there.  Any thoughts on what the syntax might be in that case?

SimenB:
https://nodejs.org/api/module.html#modulecreaterequirefilename

damdafayton:
> My workaround for this was to manually map the `uuid` module to it's commonJS export:
> 
> ```js
> // jest.config.js
> moduleNameMapper: {
[...]

PR #465, opened 2020-06-05
Export uuid as default as well

franciscop:
This allows people to do:

```js
import uuid from 'uuid';
uuid.v4();
```

It also [helps with this Jest issue](https://stackoverflow.com/a/62210156/938236), but I think allowing for the above on itself is merit enough for this PR.

### DIFF
--- a/wrapper.mjs
+++ b/wrapper.mjs
@@ -1,4 +1,7 @@
 import uuid from './dist/index.js';
+
+export default uuid;
+
 export const v1 = uuid.v1;
 export const v3 = uuid.v3;
 export const v4 = uuid.v4;


ctavan:
@franciscop thank you for the suggestion.

We deliberately removed the default export with the `uuid@7` release.

Could you elaborate on why adding a default export changes anything for jest in particular?

Isn't this more a problem of Jest not yet supporting `pkg.exports` which has been unflagged (but is still an experimental API) in Node.js 14.x?

I think we are tracking this issue in #451.

ctavan:
As noted in [stackoverflow](https://stackoverflow.com/questions/62164814/cant-use-uuid-module-when-running-tests-with-jest/62210156#comment110045407_62210156) I think adding a default export to the esm build (while the CommonJS package does not have a default export) just for the sake of working around a current flaw in jest is not justified.

I think it even has the potential of doing more harm than good because it will make testing `pkg.exports` support which is currently being developed for jest (https://github.com/facebook/jest/issues/9771) much harder.

franciscop:
@ctavan I think it's useful for people to be able to use the library either way, but if it was purposefully removed I'll close this PR.

On the jest side, it's only until they support "Package Exports", so it only fixes a temporary bug. [I explained it in detail on StackOverflow](https://stackoverflow.com/a/62210156/938236).

ctavan:
OK. I prefer to solve this on the `jest` side of things.

PR #471, opened 2020-06-22
test: remove flaky Edge 18.0 on Windows 10 test

ctavan:
Browserstack seems to have trouble accessing the local proxy on recent
Edge 18.0 / Windows 10 machines. Since Edge 18.0 is deprecated anyways
and since we still have an Edge 15.0 test as a safety net it should be
fine to just drop the Edge 18.0 test.


ctavan:
Approved through #472

PR #473, opened 2020-06-23
docs: document jspm usage

ctavan:
Thanks to pkg.exports this module is now well-supported by
https://jspm.org/ without any Node.js crypto api polyfills for the
browser.

/cc @guybedford


guybedford:
Thanks for the mention here!

PR #479, opened 2020-07-17
feat: add parse/stringify/validate/version/NIL APIs

### LINKED ISSUE #475: Support nil or empty uuid
## Is your feature request related to a problem? Please describe.

I wanted to go ahead and make a PR for adding nil uuid which is a special case. Before doing the work, I just wanted to double check to see if this has been considered before or there are any objections to it.

You can read more about nil uuid [https://en.wikipedia.org/wiki/Universally_unique_identifier#Nil_UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier#Nil_UUID)

## Describe the solution you'd like

I was thinking about adding nill to the namespace so the user can write the following code:

```
import { nil } from 'uuid';
console.log(nil()); // => '00000000-0000-0000-0000-000000000000'
```

I believe that is valid for all versions, so we dont need to implement it under v1, v2, etc.
Let me know your thoughts and I'll get to it.

### LINKED ISSUE #478: Reopened discussion: parse/unparse uuid
In the past, requests for the ability to parse/unparse UUIDs have been closed with a referral to https://www.npmjs.com/package/uuid-parse

I am aware and agree that this use case is relatively uncommon when compared to the overall use of UUIDs. However, this module is now archived with a notice that "This module is no longer maintained"

It is not even possible to file an issue or view issues on that project which makes it unfit for safe usage. My preference would be for the functionality to simply be merged into this package. However, given the strong reluctance to do that in the past, perhaps uuidjs could take over management of that module or host the same functionality in a new module?

### LINKED ISSUE #480: isUuid
I have a use case where I need to detect if something is a UUID or some other string, I know this is just a regex, but would be cool if this package just provided this.

### LINKED ISSUE #481: toBytes
we have need for decoding the standard uuid format to just the bytes, the reason for this is we want to re-encode as url encoded base64 to shorten our urls. note, if you wanted to provide a function for that as well... use something like

`base64( toBytes( uuidv4() ) )`

also an inverse

`fromBytes`

ctavan:
@broofa should we just merge this into master and release as `8.2.0`? As far as I can tell it only adds new features, no breaking changes. Or do I miss something?

Closes #475 
Closes #478 
Closes #480 
Closes #481


ctavan:
BTW we still need documentation in the README.

broofa:
@ctavban: Agreed.  As you said, there's no breaking changes.

I'll update the README.  Also, I'm going to go ahead and add a `NIL_UUID` constant as well, per discussion in #475.  In thinking further, I don't see much down side to that.

broofa:
Updated the README, but CI tests are failing.  Do you know why?  It looks like `npm run docs && git diff --quiet README.md` is erring, but when I run that command locally it succeeds.  :-/

Spent some time cleaining up the README.  Reduced the "quick start" down to just the bare minimum (making v4 uuids).  Added an API summary so people could quickly see what the new API is like.  Restyled method args, returns, and errors as tables. 

Also, made one minor functional change - `stringify()` now tests  `validate()`s generated uiud strings to ensure they're valid. (Previously you could hand it any bytes and it would blindly stringify, even if the result wouldn't pass `validate()`)

examples/benchmark/benchmark.js
```diff
@@ -5,46 +5,96 @@ const uuidv1 = (typeof window !== 'undefined' && window.uuidv1) || require('uuid
...
+      console.log(event.target.toString());
+    })
+    .on('complete', function () {
+      console.log('Fastest is ' + this.filter('fastest').map('name'));
+    })
+    .run();
+}
+
+testParseAndStringify();
+testGeneration();
```
ctavan, line 100: Will these tests run in parallel? Shouldn't we rather ensure that they are run in series?

rollup.config.js
```diff
@@ -20,4 +20,9 @@ export default [
   chunk('v3', 'uuidv3'),
   chunk('v4', 'uuidv4'),
   chunk('v5', 'uuidv5'),
+
+  chunk('version', 'uuidVersion'),
+  chunk('validate', 'uuidValidate'),
+  chunk('parse', 'uuidParse'),
+  chunk('stringify', 'uuidStringify'),
```
ctavan, line 27: I think for consistency we also need a `uuidNIL` export?!

src/stringify.js
```diff
@@ -0,0 +1,60 @@
...
+  // the input array values don't map to a defined hex octet, the string length
+  // will get blown out (e.g. "74af23d8-85undefined2c44-...")
+  //
+  // This is a somewhat crude check, but avoids having to check each value
+  // individually.
+  if (uuid.length !== 36) {
+    throw new TypeError(
+      'Stringified UUID is invalid (All input values must be integers between 0 and 255)',
+    );
+  }
```
ctavan, line 49: If we do the expensive validation afterwards this would only be a perf improvement for the case where people try to stringify malformed byte arrays. I don't think this is worth the extra code.

Maybe just keep the additional validation?

What's the perf impact of the additional validation on `v4` and `v1` UUID generation?

broofa: Good catch.  Making that change.

Regarding perf, validating in stringify does have an impact on uuid generation but, personally, not enough that I'm overly concerned.  I suppose we could add a `noValidate` arg to `stringify()` to allow callers to suppress the check, but I'm okay leaving that out for the time being and waiting until someone makes a compelling case for it.

**Perf with stringify() validation**:

```
uuidv1() x 1,613,447 ops/sec ±0.60% (94 runs sampled)
uuidv1() fill existing array x 9,143,504 ops/sec ±0.34% (99 runs sampled)
uuidv4() x 397,873 ops/sec ±1.02% (89 runs sampled)
uuidv4() fill existing array x 556,775 ops/sec ±0.59% (97 runs sampled)
uuidv3() x 212,358 ops/sec ±1.71% (88 runs sampled)
uuidv5() x 219,256 ops/sec ±1.88% (88 runs sampled)
```

**Perf w/out stringify() validation**:

```
uuidv1() x 2,183,327 ops/sec ±0.59% (90 runs sampled)
uuidv1() fill existing array x 9,118,317 ops/sec ±0.21% (93 runs sampled)
uuidv4() x 537,945 ops/sec ±0.36% (93 runs sampled)
uuidv4() fill existing array x 554,379 ops/sec ±0.37% (95 runs sampled)
uuidv3() x 257,289 ops/sec ±0.56% (83 runs sampled)
uuidv5() x 254,984 ops/sec ±2.09% (76 runs sampled)
```

ctavan:
Changes look good. What about my comment regarding the benchmark parallelism?

examples/benchmark/benchmark.js
broofa: > Will these tests run in parallel? Shouldn't we rather ensure that they are run in series?

I believe Benchmark.js runs synchronously, so these do run in series.  (At least, that's what I see when I do `cd examples/benchmark; node benchmark.js`.

zaggino:
@broofa @ctavan I upgraded package from `8.2.0` to `8.3.0` and this should be marked as a breaking change (major) update. Before the routine would work:
```
import { v5 as uuidv5 } from 'uuid';

const UUID_V5_NAMESPACE = '00112233-4455-6677-8899-aabbccddeeff';

function getUuidHash(str: string) {
  return uuidv5(str, UUID_V5_NAMESPACE);
}
```

Now after update it throws an error:
```
TypeError: Invalid UUID
```

ctavan:
@zaggino the [RFC requires the namespace to be a valid UUID](https://tools.ietf.org/html/rfc4122#section-4.3). It's true that up to `8.2.0` we did not verify that the passed namespace was a truly valid UUID, and we added this check in `8.3.0`. However this should rather be considered a bug fix than a breaking change since the behavior before was counter the RFC.

Can you update your code to pass only valid UUIDs as namespaces? A valid UUID must match this regex:

https://github.com/uuidjs/uuid/blob/ade365502540138326f9c3dcfe75147549fa9632/src/regex.js#L1

I.e. it must either be the NIL UUID or it has the UUID version after the second hyphen (in your example there's a `6` which is not a valid UUID version) and one of `89ab` after the third hypen (you have an `8` in your example here, so this would be valid).

zaggino:
I did update the code, but my concern was that a `^8.0.0` dependency which we had caused a break all over the application when a package-lock.json was refreshed.

From the RFC perspective, I agree on fixing this, but from a user-developer perspective - I never noticed any buggy behaviour before this, so was there ever an actual bug when a namespace which was not a 100% proper uuid was used?

ctavan:
I think this can be a serious interoperability issue: if you have other libraries or programming languages that already do the right thing by validating the passed namespace UUIDs then the UUIDs generated from malformed namespaces with old versions of this library won’t be compatible with those other libraries/languages.

So as long as everything that ever has to deal with the v5 UUIDs generated off of bad namespace ids will always use the same flawed library there’s probably no problem, but as soon as you need to be interoperable this might cause very subtle and hard-to-find bugs.

ctavan:
All that said I agree that it is maybe not the best developer experience to introduce such a change in a minor version bump. Unfortunately it’s out now and too late. Our assessment that this should be a non-breaking change for most developers was apparently wrong. We‘ll be more careful in the future. I’m sorry for the inconvenience.

PR #486, opened 2020-07-21
feat: load env vars from .env file (development feature)

broofa:
Enables `.env` file support for wdio.  (This lets me use my alternate browserstack account for this project, without affecting other projects).

To test: create a `.env` file in project root that looks something like this:
```
BROWSERSTACK_USER=your_user_name
BROWSERSTACK_ACCESS_KEY=your_access_key
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist/
 browserstack.err
 local.log
 *.tgz
+.env
--- a/examples/browser-rollup/package-lock.json
+++ b/examples/browser-rollup/package-lock.json
@@ -315,7 +315,17 @@
       }
     },
     "uuid": {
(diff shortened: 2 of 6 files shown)


ctavan:
I'm a huge fan of https://direnv.net/ for this purpose. I like the fact that it doesn't require any code changes. Would that be an option for you as well?

It's really helpful for other use cases. For example `uuid` is one of the few projects where I still want to run `node@12` instead of `node@14`, so I just put

```
PATH_add /usr/local/opt/node@12/bin
```

to my `.envrc` and automatically get `node@12` whenever inside the `uuid` project.

broofa:
Got direnv working... nice!

PR #489, opened 2020-07-22
test: add browser tests for new APIs

broofa:
Adds browser tests for stringify, parse, version, and validate methods.

Notes:
* The examples-all page was getting a bit long when displayed in the browser so I tweaked the structure a bit.  Specifically, I switched from H2/P tags to DT/DD tags since that seemed more semantically appropriate, and added an inline `<style>` block clean up the appearance.
* The whole "examples" directory feels a bit off to me.  All of the scaffolding we've added to allow these to be useful in testing gets in the way of providing readers with clean, easy-to-comprehend examples.   I'm wondering if we shouldn't move this stuff under the "./tests" directory, instead.
* I added new tests to browser-rollup and browser-webpack examples, since that's what's CI tested, but none of the other example directories.  Should I do that for consistency?  (And, if so, it'd be nice if that were generalized somehow to avoid having to copy/paste code across those directories.)


broofa:
BrowserStack is kickin' my ass today.  When I do `npm run test:browser` from my shell, everything passes but the BrowserStack dashboard shows the "manual build" as failing.  Conversely, the dashboard shows the latest `v9_tests` build as passing, but here in this Github PR, it says "checks were not successful".

I think the above failures are because "Terser" lint checks are failing, but I'm unable to reproduce that on my local box, nor are the error messages in the log very helpful.

.gitignore
```diff
@@ -4,3 +4,4 @@ dist/
 browserstack.err
 local.log
 *.tgz
+.envrc
```
ctavan: Preferably put this into your global `~/.gitignore` instead since this is not project-specific but specific to the fact that you are using direnv.

examples/utils/testpage.js
```diff
@@ -6,8 +6,8 @@ export default function test(callback) {
         'body {font-family: monospace;}',
         'dt, dd {display: inline-block; margin: 0;}',
         'dt {min-width: 15em;}',
-      ].join('\n'),
-    ),
+      ].join('\n')
```
ctavan, line 9: Ooof, that is a real shame that we apparently need to change the prettier setting just for this one single trailing comma 😭

.gitignore
broofa: done

ctavan:
I agree that the current setup is not ideal: After all, the examples were there first and I only started using them as test cases when I was adding browser tests. This definitely deserves a refactoring in the future, but I think we can pull this in before.

PR #490, opened 2020-07-22
chore: run npm audit fix to satisfy dependabot

ctavan:
TBH dependabot is such a shitshow. Bothering us about irrelevant minor security flaws in dev dependencies, that, upon each merge to master, cause 3 pull-requests to be auto-rebased, overwhelming our Browserstack instance, leading to timeouts for tests on master, which I then have to re-run manually.

I have disabled dependabot and I suggest to leave it disabled as long as this library doesn't have any production dependencies.

Or am I missing something?

### DIFF
--- a/examples/benchmark/package-lock.json
+++ b/examples/benchmark/package-lock.json
@@ -15,9 +15,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.15",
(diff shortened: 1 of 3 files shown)


TrySound:
Totally agree. Closing notifications about lodash upgrade made me nervous.

PR #495, opened 2020-07-27
fix: add missing exports and tests for new APIs

ctavan:
The new APIs were still missing from some environment-specific tests. We
should definitely consider consolidating these environment-specific test
cases somehow, this currently requires way too much manual work and it's
too easy to miss certain places that need an update.


ctavan:
> Agreed! It'd be nice to have a single location for this stuff. Maybe a build step of some sort that generates the source? (Seems complicated, though... not sure how I really feel about that.)

I'm also not entirely sure how a solution should look like. I have created #496 to explore this further. I think it's not super urgent but something we should attack at some point to keep this module maintainable.

> I confess to somewhat deliberately ignoring these when doing the first round of new tests. Dealing with BrowserStack was enough of a hassle I deliberately ignored this side of things. Thanks for picking these up.

I feel you 😉

PR #504, opened 2020-08-03
docs: document byte order in parse() and stringify(). fixes #503

### LINKED ISSUE #503: Unexpected/incompatible byte representation
I can roundtrip GUID `75b097d1-b891-4834-9b12-4a94d8c42504` with the following node repl code:

```node
> const uuid = require('uuid')
> uuid.parse('75b097d1-b891-4834-9b12-4a94d8c42504')
Uint8Array(16) [
  117, 176, 151, 209, 184,
  145,  72,  52, 155,  18,
   74, 148, 216, 196,  37,
    4
]
> const a = uuid.parse('75b097d1-b891-4834-9b12-4a94d8c42504')
undefined
> uuid.stringify(a)
'75b097d1-b891-4834-9b12-4a94d8c42504'
```

But I suspect the byte representation is not as expected. I would expect the byte array to be [209, 151, 176, 117, 145, 184, 52, 72, 155, 18, 74, 148, 216, 196, 37, 4]. I found a gist at https://gist.github.com/daboxu/4f1dd0a254326ac2361f8e78f89e97ae that produces that array from `75b097d1-b891-4834-9b12-4a94d8c42504` and it's also what is produced by fx. C# / .NET.

broofa:


broofa:
> Should we mention though that this is the byte order recommended by the RFC

Let's not.  Between the note and the examples, it's clear what we're doing.  I don't think we need to go into the "why" behind it.

PR #509, opened 2020-08-13
wdio is used in tests only, does not install in some platforms (z/OS) - make it optional

IgorTodorovskiIBM:
The wdio node module is not supported on os390, preventing uuid from being installed.  Since wdio is used for specific testing purposes only, I made it an optionalDependency.  That way, if wdio fails to install, uuid will still be installed successfully.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,6 @@
       "version": "7.5.5",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
       "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
-      "dev": true,
       "requires": {
         "@babel/highlight": "^7.0.0"
       }
@@ -348,7 +347,6 @@
       "version": "7.5.0",
       "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.5.0.tgz",
       "integrity": "sha512-7dV4eu9gBxoM0dAnj/BCFDW9LFU0zvTrkq0ugM7pnHEgguOEeOz1so2ZghEdzviYzQEED0r4EAgpsBChKy1TRQ==",
-      "dev": true,
       "requires": {
         "chalk": "^2.0.0",
         "esutils": "^2.0.2",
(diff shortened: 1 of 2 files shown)


ctavan:
@IgorTodorovskiIBM are you trying to develop `uuid` on os390? `wdio` is only listed in the `devDependencies`, so when using `uuid` as a regular dependency, `npm` should never try to install `wdio`. How are you trying to install `uuid` on os390?

ctavan:
And out of curiosity: Are we really talking about https://en.wikipedia.org/wiki/OS/390? 😲

IgorTodorovskiIBM:
Yeah, it's z/OS actually.   UUID part of CITGM (https://github.com/nodejs/citgm) and we're trying to get the uuid test passing.   It clones the repo, does an `npm install` and then a `npm test`.  

hmm, I see that there's no optionalDevDependencies support.

ctavan:
OK. Could you elaborate a bit more on why you need to install it that way and why you need to get the tests passing? I think a bit more context would help us to find a suitable solution.

IgorTodorovskiIBM:
Sure, we're porting Node.js to z/OS and CITGM is one group of tests that the Node community runs.  We're trying to get that to 100%.   UUID is listed in the set of tests (https://github.com/nodejs/citgm/blob/master/lib/lookup.json).

CITGM only fetches the uuid source from github and then runs `npm install` and then runs `npm test` on the module.  Since `wdio` is not used, we wanted to skip it or make it an optional dependency

ctavan:
Got it, thanks! BTW I was not aware of `uuid` being part of CITGM, learned something today 😉 

Unfortunately, putting the `wdio` dependencies into `optionalDependencies` is not really an option because that would install them as regular production dependencies.

I guess we'll have to find a different way but I think I'll have to do some research.

ctavan:
An ugly workaround that comes to my mind would be to put all `wdio` dependencies as `optionalDependencies` into a wrapper package and then reference that wrapper packages in `devDependencies` of uuid instead. That way their failing installation on z/OS might simply be ignored. Unfortunately that would also increase the maintenance burden for this module quite a bit.

Does anyone have better ideas?

ctavan:
@MylesBorins do you maybe have a good idea of how to handle this issue? I can well imagine that `uuid` might actually be used on s390 so I'd rather not just skip tests for that platform. But we indeed don't need to run any browser tests there, so some way to skip installation `@wdio` dependencies for that platform would be neat.

MylesBorins:
@ctavan The wrapper module approach is a reasonable idea... it's basically what chokidar does to make fsevents optional for file watching.

If it would take a bit to work this out we very well can skip this for 390x until it's ready

ctavan:
> @ctavan The wrapper module approach is a reasonable idea... it's basically what chokidar does to make fsevents optional for file watching.
> 
> If it would take a bit to work this out we very well can skip this for 390x until it's ready

Thanks for the quick reply!

@IgorTodorovskiIBM should comment how urgent getting the tests green is for them.

I think I can get the wrapper approach done within a couple of days.

IgorTodorovskiIBM:
> > @ctavan The wrapper module approach is a reasonable idea... it's basically what chokidar does to make fsevents optional for file watching.
> > If it would take a bit to work this out we very well can skip this for 390x until it's ready
> 
> Thanks for the quick reply!
> 
> @IgorTodorovskiIBM should comment how urgent getting the tests green is for them.
> 
> I think I can get the wrapper approach done within a couple of days.

That would be great!  It's not urgent

ctavan:
@IgorTodorovskiIBM see #510 for a potential solution.

IgorTodorovskiIBM:
Thanks, I'll close this PR

PR #517, opened 2020-09-11
chore: require minimal reproducible example for bugs

ctavan:
Since people tend to never provide minimal reproducible examples I want
to make this a strict requirement for bug reports.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -17,9 +17,7 @@ version before filing your bug report. Node.js 13.x is NOT supported by this lib
 
 ## How to reproduce
 
-[Instructions for how we (the `uuid` maintainers) can reproduce the problem]
-
-- [ ] &#X1f389; This is a "[Minimal, Complete, and Verifiable](https://stackoverflow.com/help/minimal-reproducible-example)" Example &#X1f389;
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -17,9 +17,7 @@ version before filing your bug report. Node.js 13.x is NOT supported by this lib
 
 ## How to reproduce
 
-[Instructions for how we (the `uuid` maintainers) can reproduce the problem]
-
-- [ ] &#X1f389; This is a "[Minimal, Complete, and Verifiable](https://stackoverflow.com/help/minimal-reproducible-example)" Example &#X1f389;
+[Provide a [Minimal, Complete, and Verifiable](https://stackoverflow.com/help/minimal-reproducible-example) so that we (the `uuid` maintainers) can reproduce the problem. Bug reports without minimal reproducible example may be closed without further investigations.]
```
broofa, line 20: I would be more blunt about this ...

```suggestion
**REQUIRED**: [Provide a [Minimal, Complete, and Verifiable](https://stackoverflow.com/help/minimal-reproducible-example) so that we (the `uuid` maintainers) can reproduce the problem.]  Issues that do not provide this will be closed without investigation.
```

broofa (approved):

PR #524, opened 2020-10-04
chore: upgrade all devDependencies

ctavan:


ctavan:
I merged this right away. I believe devDependency version bumps don't require review from other maintainers. Please let me know if you think otherwise.

PR #529, opened 2020-11-01
fix: add crypto at dependencies

SoYoung210:
## Description

Webpack5 does not automatically install crypto, and an error occurs when building a package that uses uuid.

Since [src/rng.js](https://github.com/uuidjs/uuid/blob/252ebcfb33b2d056963a530fc02067ae3de6095a/src/rng.js#L1) is using crypto, I think it needs to be added to the dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5115,6 +5115,11 @@
         "which": "^1.2.9"
       }
     },
+    "crypto": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/crypto/-/crypto-1.0.1.tgz",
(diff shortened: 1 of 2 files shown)


ctavan:
`crypto` is not a dependency of this library.

* When bundling for Node.js, this library will use the Node.js core `crypto`-library.
* When bundling for the Browser, this library will use WebCrypto available in all modern browsers.

My guess is that something is odd with your webpack config and webpack somehow processes the files dedicated for Node.js where it should use those dedicated for the browser. Feel free to have a look at the examples for [Node.js](https://github.com/uuidjs/uuid/tree/master/examples/node-webpack) and [Browser](https://github.com/uuidjs/uuid/tree/master/examples/browser-webpack) webpack builds for a working example, maybe something around module resolution and the [`resolve.mainFields`](https://webpack.js.org/configuration/resolve/#resolvemainfields) webpack option.

Issue #498, opened 2020-07-28
Provide regular expressions and JSON schemas for UUIDs


goloroden:
Hi 👋 

I'm one of the authors of the [uuidv4](https://www.npmjs.com/package/uuidv4) module.

With the new 8.3 version of `uuid` there is almost everything there that is needed from our POV, and we are planning to deprecate `uuidv4` in favor of this module.

However, there are two very small things missing, before we can do this:

First, it would be awesome to have regular expressions for UUIDs.
In `uuidv4` we use the following ones:

```javascript
const regex = {
  v4: /(?:^[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[a-f0-9]{4}-[a-f0-9]{12}$)|(?:^0{8}-0{4}-0{4}-0{4}-0{12}$)/u,
  v5: /(?:^[a-f0-9]{8}-[a-f0-9]{4}-5[a-f0-9]{3}-[a-f0-9]{4}-[a-f0-9]{12}$)|(?:^0{8}-0{4}-0{4}-0{4}-0{12}$)/u
};
```

Besides, it would be great to have JSON schemas for UUIDs. Again, we're using this code here, which builds on top of the aforementioned regexs:

```javascript
const jsonSchema = {
  v4: { type: 'string', pattern: regex.v4.toString().slice(1, -2) },
  v5: { type: 'string', pattern: regex.v5.toString().slice(1, -2) }
};
```

Unfortunately, we are missing the time to update the `uuid` module ourselves, however this should be fairly easy to integrate.

ctavan:
Thank you for coming back here @goloroden! We're definitely interested in covering your use cases in order to reduce the number of `uuid` libraries in the JS ecosystem.

To make progress here I would like to understand your use cases for the above two features a bit better.

Regarding the regexes: Applying e.g. the `v4` regex would be more or less equivalent to the following code:

```js
import { version } from 'uuid';

let isV4 = false;
try {
  isV4 = version('SOME_UUID') === 4;
} catch (e) {}
```

If you want to avoid the error handling you could do:

```js
import { validate, version } from 'uuid';

const isV4 = validate(SOME_UUID) && version(SOME_UUID) === 4;
```

That said we have discussed exposing the plain regex(es) in the past so this might definitely be an option, I would just like to understand the benefits over the existing APIs like `validate()` and `version()`.

[...]

PR #534, opened 2020-11-19
feat: exports regex as 1st class citizen

joaomilho:
This will help the following use case (and others):

```js
import { regex } from 'uuid';

describe('duck', () => {
  it('quacks', () => {
   const duck = createDuck();
    expect(duck).toEqual({
      name: 'daffy',
      sound: 'quack',
      uuid: expect.stringMatching(regex)
    })
  })
})
```

Importing from dist/esm is not ideal, and proper TS support (I think we'll have to change @types/uuid for this) would be great.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,3 +7,4 @@ export { default as version } from './version.js';
 export { default as validate } from './validate.js';
 export { default as stringify } from './stringify.js';
 export { default as parse } from './parse.js';
+export { default as regex } from './regex.js';


broofa:
See https://github.com/uuidjs/uuid/issues/498#issuecomment-665163121

tl;dr: Allowing users to depend on the regex we use for validate() makes it difficult to modify if/when we need to do so in the future.

Doesn't seem like it's that hard to workaround ...

```javascript
expect({...duck, uuid: uuid.validate(duck.uuid)}).toEqual({
  name: 'daffy',
  sound: 'quack',
  uuid: true
})
```

Or...
```javascript
expect(uuid.validate(duck.uuid)).toBe(true);
expect(duck).toEqual({
   name: 'daffy',
   sound: 'quack',
   uuid: expect.any(String)
});
```

broofa:
Closing due to lack of activity.

joaomilho:
Fair enough. But perhaps a uuid matcher in a diff lib?

```js
describe('duck', () => {
  it('quacks', () => {
   const duck = createDuck();
    expect(duck).toEqual({
      name: 'daffy',
      sound: 'quack',
      uuid: expect.stringMatchingUUIDV4()
    })
  })
})
```

piranna:
> perhaps a uuid matcher in a diff lib?

I like this :-)

Issue #536, opened 2020-11-20
uuid should not rely on module load order


drewthaler:
## Is your feature request related to a problem? Please describe.

uuid 7.x uses Crypto.getRandomValues which is not implemented in react-native, and attempts to work around that by asking clients to require another module — react-native-get-random-values — to implement a polyfill. It further caches the polyfill at module load time, requiring users to enforce module load order or else it breaks.

First, relying on module load order is _fragile_ and prone to breakage. In particular, I've run into a problem with Expo where the expo-updates Updates.reloadAsync is reloading our app code in a situation where the uuid module is already initialized — probably because expo-updates uses it. 

Second, this causes confusion among developers. At present it's ~10% of the last 32 github issues. (#531, #514, #499). 

## Describe the solution you'd like

Multiple possible solutions could be implemented. They are not mutually exclusive.

1. uuid could **directly** rather than indirectly depend on this module that it certainly depends on, importing a symbol and using that, rather than hoping the polyfill gets installed in time. 
2. uuid could fold a minimal version of the polyfill code (which is rather simple) into the module and use that as a fallback
3. uuid could lazy-load the getRandomValues symbol on invocation, rather than caching it at module load time. Then at least you just need the polyfill present by the point of use. 

## Describe alternatives you've considered

See above list. Option 1 seems like the best choice to me, though option 2 would at least restore the pre-7.x functionality. The other two are mainly workarounds for not doing the first one. Relying on module load order is really the core issue, and until uuid stops relying on module load order it will continue to have this problem.

ctavan:
@drewthaler thank you for raising this. However there are several reasons why we do it the way we do it:

> 1. uuid could **directly** rather than indirectly depend on this module that it certainly depends on, importing a symbol and using that, rather than hoping the polyfill gets installed in time.

The biggest use case for this library is still with Node.js and not React Native. Adding a React Native specific polyfill as a dependency in `package.json` is not an option.

> 2. uuid could fold a minimal version of the polyfill code (which is rather simple) into the module and use that as a fallback

It was a deliberate decision to drop the insecure random number generator fallback that was present up until `uuid@3.x`, feel free to read through #173 and related issues (like #354).

> 3. uuid could lazy-load the getRandomValues symbol on invocation, rather than caching it at module load time. Then at least you just need the polyfill present by the point of use.

[...]

PR #537, opened 2020-11-20
lazy-load getRandomValues to mitigate problems with react-native-get-random-values polyfill

drewthaler:
Here's a PR for the lazy loading fix discussed in #536. This defers the loading and caching of crypto.getRandomValues to first invocation, rather than module load time, helping out React Native clients.

Without this change, RN projects must manage an implicit dependency on react-native-get-random-values, and try to ensure that it gets loaded before uuid does.

With this change, the RN polyfill simply needs to be installed sometime before the first uuid is created.

No significant benchmark differences. (In fact the v4 tests ran slightly faster on my machine with this change, but I'm sure that was accidental.)

### DIFF
--- a/src/rng-browser.js
+++ b/src/rng-browser.js
@@ -2,23 +2,27 @@
 // require the crypto API and do not support built-in fallback to lower quality random number
 // generators (like Math.random()).
 
-// getRandomValues needs to be invoked in a context where "this" is a Crypto implementation. Also,
-// find the complete implementation of crypto (msCrypto) on IE11.
-const getRandomValues =
-  (typeof crypto !== 'undefined' &&
-    crypto.getRandomValues &&
-    crypto.getRandomValues.bind(crypto)) ||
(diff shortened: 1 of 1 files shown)


ctavan (approved):
Thank you for the PR and thank you for even running the benchmarks! A v4 difference is certainly coincidence since this code is not executed in the v4 benchmark.

@broofa fine with this as well?

TrySound (approved):

broofa (approved):

ctavan:
@drewthaler I've published `uuid@8.3.2-beta`. Would you mind testing if the polyfill import order is now fixed for your use case?

ctavan:
As reported in https://github.com/uuidjs/uuid/issues/536#issuecomment-739085989 this seems to work.

Released as `uuid@8.3.2`. 🎉 

Thanks again for the contribution 🙇

PR #543, opened 2020-12-07
docs: add examples of per-version validation

Mearman:
Created in lieu of specific per-version validation methods as seen in https://github.com/uuidjs/uuid/pull/542

Might be something stupid I'm doing but for some reason, having the two aliased imports caused `SyntaxError: Cannot use import statement outside a module`

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,23 @@ uuidValidate('not a UUID'); // ⇨ false
 uuidValidate('6ec0bd7f-11c0-43da-975e-2a8ad9ebae0b'); // ⇨ true
 ```
 
+Using `validate` and `version` together it is possible to do per-version validation, e.g. validate for only v4 UUIds.
+
+```javascript
+import { version as uuidVersion } from 'uuid';
+import { validate as uuidValidate } from 'uuid';
+
+function uuidValidateV4(uuid) {
+  return uuidValidate(uuid) && uuidVersion(uuid) === 4;
+}
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -296,6 +296,23 @@ uuidValidate('not a UUID'); // ⇨ false
 uuidValidate('6ec0bd7f-11c0-43da-975e-2a8ad9ebae0b'); // ⇨ true
 ```
 
+Using `validate` and `version` together it is possible to do per-version validation, e.g. validate for only v4 UUIds.
+
+```javascript
+import { version as uuidVersion } from 'uuid';
+import { validate as uuidValidate } from 'uuid';
```
ctavan, line 303: I guess runmd stumbles when you put both imports into one line? Because we actually just [hack our way around ESM](https://github.com/broofa/runmd/commit/cdeb284a3dc3c3b48b6626167058a2ea68987f7a)…

ctavan (approved):
Apart from putting both imports into one statement (which is probably currently not supported by runmd) I really like this! Thank you 🙇‍♂️

README.md
Mearman: Hahah yeah it wasn't happy about it at all

broofa: > Want me to submit it as an issue over on runmd?

Yes, please.   I'm aware of the issue but need to revisit it to see if there's a better solution than the one ctavan and I hacked in a while ago.

broofa (approved):
Looks good, thank!

PR #545, opened 2021-01-04
Change default export to named function

dylans:
Frameworks like Next.js complain about anonymous export default functions, e.g.

`Anonymous function declarations cause Fast Refresh to not preserve local component state.`

Versions beside 3.5 already use named exports, so update 3.5 to do the same.

### DIFF
--- a/src/v35.js
+++ b/src/v35.js
@@ -16,7 +16,7 @@ function stringToBytes(str) {
 export const DNS = '6ba7b810-9dad-11d1-80b4-00c04fd430c8';
 export const URL = '6ba7b811-9dad-11d1-80b4-00c04fd430c8';
 
-export default function (name, version, hashfunc) {
+function v35(name, version, hashfunc) {
   function generateUUID(value, namespace, buf, offset) {
     if (typeof value === 'string') {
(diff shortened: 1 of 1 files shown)


src/v35.js
```diff
@@ -16,7 +16,7 @@ function stringToBytes(str) {
 export const DNS = '6ba7b810-9dad-11d1-80b4-00c04fd430c8';
 export const URL = '6ba7b811-9dad-11d1-80b4-00c04fd430c8';
 
-export default function (name, version, hashfunc) {
+function v35(name, version, hashfunc) {
```
ctavan, line 19: Why not just `export default function v35(…`?

ctavan (changes requested):
Interesting, thank you for submitting a patch!

I think it can be slightly simplified, see inline comment. Would you mind updating the PR?

src/v35.js
dylans: I actually prefer this, I just went with matching what was in src/v4.js 👍

TrySound:
Btw named exports does not allow anonymous functions 😁

ctavan (approved):
LGTM

PR #553, opened 2021-01-30
docs: add command 'npx' for CLI usage

pandori3:
It's very convenient and wonderful to be able to use uuid generation on the command line!
I think that it is convenient to use in npx.

### Before
```
$ uuid
ddeb27fb-d9a0-4624-be4d-4615062daed4
```

### After
```
$ npx uuid
ddeb27fb-d9a0-4624-be4d-4615062daed4
```


ctavan:
In fact, this is how I use the `uuid` cli all the time as well. I think it's worth presenting it like that in the readme as well.

PR #563, opened 2021-03-04
Change default export

### LINKED ISSUE #561: No default export provided
## Describe the bug

Many packages import uuid as:

```js
const uuid = require('uuid');
```

When converted into ES modules, this would correspond to:

```js
import uuid from 'uuid';
```

When running the browser though, the browser ES module build does not support a "default" export and as a result packages on jspm using UUID are currently breaking.

## How to reproduce

Try, for example, `import('https://jspm.dev/@google-cloud/bigquery')`

## Expected behavior

UUID should be configured such that the shape of the module between browsers and Node.js remains the same - the default export should exist for both. Node.js CommonJS modules are always represented by a default export, see https://nodejs.org/dist/latest-v15.x/docs/api/esm.html#esm_commonjs_namespaces for more info.

## Additional information

The suggested fix would be to add an `export default uuid` to the ES module namespace.

laughedelic:
I just applied the diff suggested by @TrySound in https://github.com/uuidjs/uuid/issues/561#issuecomment-787435073.

Hopefully it fixes #561. Let me know if this is not enough and something else has to be done. I didn't understand if https://github.com/uuidjs/uuid/issues/561#issuecomment-787438363 was another part of the solution or an alternative approach.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
         "module": "./dist/esm-browser/index.js",
         "require": "./dist/browser/index.js",
         "import": "./dist/browser/wrapper.mjs"
-      },
+      }
     },
     "./package.json": "./package.json"
   },
--- a/package.json
+++ b/package.json
@@ -25,7 +25,11 @@
         "require": "./dist/index.js",
         "import": "./wrapper.mjs"
       },
-      "default": "./dist/esm-browser/index.js"
+      "default": {
+        "module": "./dist/esm-browser/index.js",
+        "require": "./dist/browser/index.js",
+        "import": "./dist/browser/wrapper.mjs"
+      },
     },
     "./package.json": "./package.json"
   },


tchakabam:
is this another possible fix for https://stackoverflow.com/questions/62549457/error-err-package-path-not-exported-package-subpath-v4-is-not-defined-by/69564152#69564152 ?

also, see my PR https://github.com/uuidjs/uuid/pull/584

ctavan:
Is this still an issue?

ctavan:
Given this comment https://github.com/uuidjs/uuid/issues/561#issuecomment-823487310 I'm wondering if this is still an issue in any existing bundler out there?

Please feel free to reopen this PR or open a new one if you still encounter an issue with default exports.

PR #567, opened 2021-04-09
fix: upgrading from uuid3 broken link

watadarkstar:
Updates the README as it had a broken link with an extra `x` at the end.


broofa:
Good catch, and thanks for the PR.  However, [as noted](https://github.com/uuidjs/uuid/blame/master/README.md#L2) changes to the README should be made to `README_js.md`, then `npm run docs` to regenerate `README.md`.

watadarkstar:
@broofa Thanks for the note. I fixed that issue and opened a new PR: https://github.com/uuidjs/uuid/pull/568

PR #585, opened 2021-10-23
update all dependencies to @latest, drop node@8 CI testing, add node@16

broofa:
Got paranoid after seeing [this news](https://www.reddit.com/r/programming/comments/qdlela/breaking_npm_package_uaparserjs_with_more_than_7m/)  and decided to update all our dependencies to the latest version of things. 

`npm run build` and `npm test` appear to work.

Edit: `npm run lint`  broke.  Fixing this involved switching to `@eslint/babel-parser`, and then `eslint --fix` and `prettier -w` to patch formatting differences.

... and apparently `jest@latest` doesn't work with `node@8`.    So I've updated our CI platforms to drop node 8 and add node 16.

Note: I don't know why Github is saying it needs the "ci (8.x) Expected" check here, since I removed node 8 from the `node-versions` in the ci.yml workflow. 😕


broofa:
Okay, removed node 8 from checks.  Also renamed `master` to `main`.

broofa:
@ctavan ping

ctavan (changes requested):
I'm generally fine with this change, but please make sure that when you squash and merge, the resulting commit message follows https://www.conventionalcommits.org/en/v1.0.0/ to produce a breaking change entry in the changelog (e.g. like https://github.com/uuidjs/uuid/commit/daf72b84ceb20272a81bb5fbddb05dd95922cbba).

I also noticed that the README still needs an update in the summary where it still mentions Node 8.

Issue #598, opened 2021-11-25
New version release plan?


HyunSeob:
It's been almost a year since the last release, so I wonder when the new version will be released. Is there any plan for release?

ctavan:
This library is very solid and generally a rather complete implementation of RFC 4122.

What features or improvements would you be looking for in a new release?

HyunSeob:
Thank you for the response!

When I use this package with my Next.js project, the warning message like `Anonymous function declarations cause Fast Refresh to not preserve local component state.` is printed out. So I think it would be nice that PR #545 is released.

ctavan:
Thanks for pointing this out again. Looks like we should make a new release soon then.

@broofa do you think we should craft a patched 8.x release that contains the specific bug fix or would you go for 9.0 directly (as node 8 support was dropped)?

Maybe for 9.0 we could create a list of things we want to include (like dropping IE 11 support), to not have too many major version bumps?

broofa:
Let's roll a 9.0 release.  Enough [stuff has changed](https://github.com/uuidjs/uuid/compare/v8.3.2...main).  In addition to already-committed work, let's include the following:
* [ ] #597
* [ ] #590
* [ ] #469

Rush:
Maybe you could pull the types directly into this repo? https://www.npmjs.com/package/@types/uuid

ctavan:
> Maybe you could pull the types directly into this repo? https://www.npmjs.com/package/@types/uuid

According to the typescript documentation this is still not recommended for packages that don’t auto-generate the typescript declarations: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

I would prefer following the official typescript guidance and keep the types separate.

TrySound:
This is usually meant for those who add types and does not maintain them. For consumer this is burden to add one more package with possibly outdated types (this is the feeling I always have about any `@types/*` package).

For this project I don't see an issue in improving DX.

ctavan:
> This is usually meant for those who add types and does not maintain them. For consumer this is burden to add one more package with possibly outdated types (this is the feeling I always have about any `@types/*` package).
> 
> For this project I don't see an issue in improving DX.

I.e. you would prefer to have the types shipped directly with the uuid package?

[...]

PR #600, opened 2021-11-29
feat: use native crypto.randomUUID when available

LinusU:
*Before:*

```
uuidv4() x 2,232,325 ops/sec ±0.12% (95 runs sampled)
uuidv4() fill existing array x 8,398,262 ops/sec ±0.16% (97 runs sampled)
```

*After:*

```text
uuidv4() x 25,076,827 ops/sec ±1.93% (96 runs sampled)
uuidv4() fill existing array x 8,407,653 ops/sec ±0.21% (97 runs sampled)
```

This uses the native `crypto.randomUUID` if available, which results in more than an **11x** speedup on Node.js v16.11.1 on my M1 MacBook Air. All this while only increasing the bundle size with **34 bytes**!

Screenshot 2021-11-30 at 09 46 51

The patch implements my changes described here: https://github.com/uuidjs/uuid/pull/555#issuecomment-770720130

Closes #555

### DIFF
--- a/bundlewatch.config.json
+++ b/bundlewatch.config.json
@@ -5,9 +5,9 @@
     { "path": "./examples/browser-rollup/dist/v4-size.js", "maxSize": "0.7 kB" },
     { "path": "./examples/browser-rollup/dist/v5-size.js", "maxSize": "1.5 kB" },
 
-    { "path": "./examples/browser-webpack/dist/v1-size.js", "maxSize": "1.3 kB" },
-    { "path": "./examples/browser-webpack/dist/v3-size.js", "maxSize": "2.5 kB" },
-    { "path": "./examples/browser-webpack/dist/v4-size.js", "maxSize": "1.0 kB" },
-    { "path": "./examples/browser-webpack/dist/v5-size.js", "maxSize": "1.9 kB" }
+    { "path": "./examples/browser-webpack/dist/v1-size.js", "maxSize": "1.0 kB" },
+    { "path": "./examples/browser-webpack/dist/v3-size.js", "maxSize": "2.1 kB" },
+    { "path": "./examples/browser-webpack/dist/v4-size.js", "maxSize": "0.7 kB" },
+    { "path": "./examples/browser-webpack/dist/v5-size.js", "maxSize": "1.5 kB" }
   ]
 }
--- a/src/v4.js
+++ b/src/v4.js
@@ -3,11 +3,11 @@ import rng from './rng.js';
 import stringify from './stringify.js';
 
 function v4(options, buf, offset) {
-  options = options || {};
-
-  if (native.randomUUID && buf == null && options.random == null && options.rng == null) {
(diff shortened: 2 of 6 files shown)


src/v4.js
```diff
@@ -1,9 +1,14 @@
...
 import rng from './rng.js';
 import stringify from './stringify.js';
 
 function v4(options, buf, offset) {
   options = options || {};
 
+  if (native.randomUUID && buf == null && options.random == null && options.rng == null) {
+    return native.randomUUID();
+  }
+
```
broofa, line 11: Maybe avoid inspecting what's inside `options` and just do this ...

```suggestion
  if (native.randomUUID && buf == null && !options) {
    return native.randomUUID();
  }

  options = options || {};

```

broofa (changes requested):
One minor change, otherwise looks good.  This will fit in well with the other changes planned for #598

ctavan (approved):
Thank you for this patch, that’s awesome! Special thanks for covering this with a test case.

When merging, please make sure to craft a comment message that will result in a meaningful changelog entry.

src/v4.js
LinusU: Nice 👍 

Also changed `buf == null` to `!buf` in order to exactly match the `if` statement further down, and also shave of a few more bytes from the bundle ☺️

bundlewatch.config.json
```diff
@@ -5,9 +5,9 @@
...
     { "path": "./examples/browser-rollup/dist/v5-size.js", "maxSize": "1.5 kB" },
 
-    { "path": "./examples/browser-webpack/dist/v1-size.js", "maxSize": "1.3 kB" },
-    { "path": "./examples/browser-webpack/dist/v3-size.js", "maxSize": "2.5 kB" },
-    { "path": "./examples/browser-webpack/dist/v4-size.js", "maxSize": "1.0 kB" },
-    { "path": "./examples/browser-webpack/dist/v5-size.js", "maxSize": "1.9 kB" }
+    { "path": "./examples/browser-webpack/dist/v1-size.js", "maxSize": "1.0 kB" },
+    { "path": "./examples/browser-webpack/dist/v3-size.js", "maxSize": "2.1 kB" },
+    { "path": "./examples/browser-webpack/dist/v4-size.js", "maxSize": "0.7 kB" },
+    { "path": "./examples/browser-webpack/dist/v5-size.js", "maxSize": "1.5 kB" }
```
LinusU, line 11: Not strictly related to this PR but I noticed since I was working a lot with bundlewatch. It seems like WebPack got better with time and now matches the output size that Rollup gives us, so lowered these values to be the same for both.

LinusU:
@broofa I've adressed your comment and also updated the original post to reflect the new values (same perf, even smaller bundle size!)

bundlewatch.config.json
ctavan: Nice!

broofa (approved):

PR #601, opened 2021-11-30
chore: use Node.js 16.x for building package and tests

ctavan:
Previously we were requiring Node 12.x to get somewhat reproducible npm builds. This now upgrades the requirement to the latests LTS version 16.x


TrySound (approved):

.github/workflows/ci.yml
```diff
@@ -31,14 +31,14 @@ jobs:
...
-      if: matrix.node-version == '12.x'
+      if: matrix.node-version == '16.x'
       env:
         BUNDLEWATCH_GITHUB_TOKEN: ${{ secrets.BUNDLEWATCH_GITHUB_TOKEN }}
     - run: npm run test:node
-      if: matrix.node-version == '12.x' || matrix.node-version == '14.x'
+      if: matrix.node-version == '16.x' || matrix.node-version == '14.x'
     - run: npm run test:pack
-      if: matrix.node-version == '14.x'
+      if: matrix.node-version == '16.x' || matrix.node-version == '14.x'
```
LinusU, line 44: I think that these two tests should run on `>= 12`? That is, all that supports ESM

LinusU:
> npm ERR! code EINVALIDPACKAGENAME
> npm ERR! Invalid package name ".local": name cannot start with a period

Ran into this locally when using npm 7 as well, I didn't have time to look into why there was a `.local` folder though...

ctavan:
> > npm ERR! code EINVALIDPACKAGENAME
> > npm ERR! Invalid package name ".local": name cannot start with a period
> 
> Ran into this locally when using npm 7 as well, I didn't have time to look into why there was a `.local` folder though. It's a little annoying as it makes all files show up twice when searching in VS Code 😅

The `.local` symlink is a workaround to make local file references work in the examples folder (so you can actually test local changes)… The error you mentioned should now be fixed. Not sure about vscode. Could it be added to some ignore file for the editor?

ctavan:
I had to upgrade @wdio v6 -> v7 in order to get it to run with Node 16 and unfortunately that was a bit of a rabbit hole but it should be good again :).

I couldn't resist upgrading all other build deps as well.

.github/workflows/ci.yml
ctavan: Early versions of 12.x didn't support ESM IIRC… That's why historically we didn't test there.

But I guess nowadays it's safe to run these tests on Node 12+, so thanks for the catch!

.local/uuid/abc.js
```diff
@@ -0,0 +1 @@
+sdfkl;
```
LinusU, line 1: Should this be committed?

ctavan: Oops, of course not…. I was Troubleshooting eslint/prettier… will remove

LinusU (approved):
Nice! 👍

wdio.conf.js
```diff
@@ -7,94 +7,111 @@ const GITHUB_REF = process.env.GITHUB_REF || '';
...
-    browser_version: '11.0',
-    os: 'Windows',
-    os_version: '7',
+    browserVersion: '11.0',
   },
 
   // Chrome
   {
-    ...commonCapabilities,
+    'bstack:options': {
```
LinusU, line 39: Latest `main` have another entry here, just adding this comment so we don't miss updating that when merging/rebasing. I think that the merge/rebase will apply cleanly, but there will be one entry not updated to the new format ☺️

LinusU (approved):
Neat! 🙌

broofa (approved):

PR #604, opened 2021-12-02
Drop IE 11 support

### LINKED ISSUE #469: Drop IE11 support
I believe we should drop IE11 support with the next major release. This would allow us to produce much more modern bundles.

As of June 2020 IE11 market share is only 1.4% according to https://gs.statcounter.com/

Users who really need to support IE11 would still be able to transpile our bundles or add required polyfills.

As part of this we can remove `output.ecmaVersion: 5` from the webpack configs.

ctavan:
Closes #469

After adjusting the babel preset-env targets the browser build only changed with respect to `let/const` vs. `var` usage, see [this diff](https://gist.github.com/ctavan/519360f8777905838a9398b534e530ad#file-browser-build-diff).

It all still works in IE11 🤷‍♂️ .

I could break IE11 support by removing the `msCrypto` special case in: https://github.com/uuidjs/uuid/blob/c9e076c852edad7e9a06baaa1d148cf4eda6c6c4/src/rng-browser.js#L18-L20

On the other hand those 3 lines don't really hurt…

So the question is now whether we want to be even more aggressive with respect to browser support?

The next step we could go would be to allow function default parameters (equivalent to moving `firefox 51 -> 53` and `edge 14 -> 15`), which would add [this diff](https://gist.github.com/ctavan/519360f8777905838a9398b534e530ad#file-diff-ff53-edge-15-diff).

After that, the browser code would be an untranspiled copy of what we currently have in `src/`.

So I guess this all depends on how much "modern" syntax we're planning to use in the future and what general browser support matrix we're planning for?

Looking forward to hearing your thoughts!

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ For the creation of [RFC4122](https://www.ietf.org/rfc/rfc4122.txt) UUIDs
 - **Cross-platform** - Support for ...
   - CommonJS, [ECMAScript Modules](#ecmascript-modules) and [CDN builds](#cdn-builds)
   - Node 10, 12, 14, 16
-  - Chrome, Safari, Firefox, Edge, IE 11 browsers
+  - Chrome, Safari, Firefox, Edge browsers
   - Webpack and rollup.js module bundlers
   - [React Native / Expo](#react-native--expo)
 - **Secure** - Cryptographically-strong random values
@@ -416,6 +416,10 @@ Note: If you are using Expo, you must be using at least `react-native-get-random
 ### Web Workers / Service Workers (Edge <= 18)
 
 [In Edge <= 18, Web Crypto is not supported in Web Workers or Service Workers](https://caniuse.com/#feat=cryptography) and we are not aware of a polyfill (let us know if you find one, please).
+
+### IE 11 (Internet Explorer)
+
+Support for IE11 and other legacy browsers has been dropped as of `uuid@9`. If you need to support legacy browsers, you can always transpile the uuid module source yourself (e.g. using [Babel](https://babeljs.io/)).
 
 ## Upgrading From `uuid@7`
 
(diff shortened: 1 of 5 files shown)


ctavan:
The browser tests (in IE) still pass, so something is clearly wrong 😂 … I'll have a look.

broofa (approved):
Need to remove IE 11 from README verbiage.

broofa:
Tests failing (and not because I changed the README, I swear!)

ctavan:
I'm still working out some details here… Let's get #601 and #603 merged first.

broofa:
> I'm still working out some details here… Let's get #601 and #603 merged first.

FYI, I updated the "merge into" field on these three PRs to reflect the branch ancestry.  (Makes them easier to review.)  I hope that's okay.

broofa (approved):

ctavan:
After adjusting the babel preset-env targets the browser build only changed with respect to `let/const` vs. `var` usage, see [this diff](https://gist.github.com/ctavan/519360f8777905838a9398b534e530ad#file-browser-build-diff).

It all still works in IE11 🤷‍♂️ .

I could break IE11 support by removing the `msCrypto` special case in: https://github.com/uuidjs/uuid/blob/c9e076c852edad7e9a06baaa1d148cf4eda6c6c4/src/rng-browser.js#L18-L20

On the other hand those 3 lines don't really hurt…

So the question is now whether we want to be even more aggressive with respect to browser support?

The next step we could go would be to allow function default parameters (equivalent to moving `firefox 51 -> 53` and `edge 14 -> 15`), which would add [this diff](https://gist.github.com/ctavan/519360f8777905838a9398b534e530ad#file-diff-ff53-edge-15-diff).

After that, the browser code would be an untranspiled copy of what we currently have in `src/`.

So I guess this all depends on how much "modern" syntax we're planning to use in the future and what general browser support matrix we're planning for?

Looking forward to hearing your thoughts!

ctavan:
> FYI, I updated the "merge into" field on these three PRs to reflect the branch ancestry. (Makes them easier to review.) I hope that's okay.

Thank you, working at a company that doesn't use github for almost a year, I already forgot about this handy feature 🙄 …

LinusU (approved):
Neat! 👍

broofa (approved):

broofa:
Where are we at with this?  601 and 603 have been merged.

FWIW, IE is at 0.29% market share (https://gs.statcounter.com/browser-market-share)

PR #609, opened 2022-01-04
Mark `./dist/esm-browser/*.js`  as ESM modules

shrinktofit:
See [packagejson-and-file-extensions](https://nodejs.org/api/packages.html#packagejson-and-file-extensions). This trick was also used by rollup distribution.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -16,6 +16,7 @@ babel --env-name commonjs src --source-root src --out-dir "$DIR" --copy-files --
 
 # Transpile ESM versions of files for the browser
 babel --env-name esmBrowser src --source-root src --out-dir "$DIR/esm-browser" --copy-files --quiet
+echo "{ \"type\": \"module\" }" > "$DIR/esm-browser/package.json"
 
 # Transpile ESM versions of files for node
 babel --env-name esmNode src --source-root src --out-dir "$DIR/esm-node" --copy-files --quiet


ctavan:
Thank you for raising this.

What exact problem is this solving though? You are referencing Node.js documentation, but adding a `package.json` to the `esm-browser` build subdirectory? How does this relate?

shrinktofit:
> 

In term of Node.js, `.js` files are considered as CommonJS modules except for the nearest enclosing `package.json`  contain a `"type": "module"` field.

This package declares its modules by `exports` field: https://github.com/uuidjs/uuid/blob/main/package.json#L21 . That means an import statement like `import { v4 } from 'uuid';` with `"default"` condition resolves to `node_modules/uuid/dist/esm-browser/index.js`, which however, should be ESM modules.

ctavan:
As far as I understand, Node.js will resolve what's defined inside the [`"node"`](https://github.com/uuidjs/uuid/blob/3a033f6bab6bb3780ece6d645b902548043280bc/package.json#L23-L27) block of `pkg.exports`, i.e. Node.js will never resolve to `node_modules/uuid/dist/esm-browser/index.js`.

Can you provide a [Minimal, Complete, and Verifiable](https://stackoverflow.com/help/minimal-reproducible-example) example so that we (the `uuid` maintainers) can reproduce the problem?

shrinktofit:
> 

Our company, and some bundle tools like [rollup](https://github.com/rollup/plugins/tree/master/packages/node-resolve/#exportconditions) respect to `exports` but don't have `node` condition. So it turns to use `"default": "./dist/"esm-browser/*.js"`. But any `*.js` file is treated as CommonJS module according to Node spec.

salesh:
```
```
In Angular 13 - sometimes I get a warning / sometimes not, could it be that is related to this issue?

```
Warning: ***/web/src/app/shared/***/***.component.ts depends on 'uuid'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies
```

broofa:
Closing.  Will reopen if/when the requested example is provided.

PR #612, opened 2022-01-17
Fix import versions direclty uuid

shahen94:
**Hubspot** uses **request**
**request** uses **uuid**

Which causes the following issue
Screen Shot 2022-01-17 at 21 28 04

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,13 @@
       },
       "default": "./dist/esm-browser/index.js"
     },
-    "./package.json": "./package.json"
+    "./package.json": "./package.json",
+    "./v1": "./dist/v1.js",
+    "./v2": "./dist/v2.js",
+    "./v3": "./dist/v3.js",
+    "./v4": "./dist/v4.js",
+    "./v5": "./dist/v5.js",
+    "./v35": "./dist/v35.js"
   },
   "module": "./dist/esm-node/index.js",
   "browser": {


LinusU:
@shahen94 which version of `request` are you using? Can you check with `npm ls request`?

broofa:
@shahen94  Please also do `npm ls uuid`, as I suspect this is a bundler issue.    `request`  has been deprecated since Feb 2020 and is no longer maintained.  It depends on an old version of this module (`uuid@3.4.0`) which _does_ allow for direct import of `v4`.  However we deprecated direct imports subsequent to that version.

If `request` is trying to load `v4` from a newer version of this module then I suspect there's something wrong with your `node_modules` installation, or with how your bundler/node is resolving modules.

shahen94:
@broofa @LinusU Just runned `npm ls request`

`uuid@3.4.0`

broofa:
> So the only solution that I've found is to update uuid/package.json and mark vX exportable.

So you're editing the package.json at `./node_modules/uuid/package.json`.  I.e the package.json for the [very old] `uuid~3.4.0` module?

If so, then please note we're now on version 8.x, which is the version you're asking us to patch with this PR, and which is  a very different codebase than v3.  Even if we were to do a maintenance release for version 3.x, you'd have to get `request` to pick up that change in their dependencies to fix this issue.  But that's never going to happen because `request` is frozen and no longer maintained.

I'm actually still puzzled why you're seeing this problem.  It's not like there aren't 100K's of users out there that depend on `request` without issue.  What bundle/packager toolchain are you using?

shahen94:
Hi @broofa sorry for the delayed reply.

We're using webpack for this. if we'll not merge this PR, is there any way that I can monkey-patch the package locally ?

TrySound:
This problem may appear if you configure `resolve.modules` incorrectly.

TrySound:
This will let you patch any package in your node_modules on postinstall
https://github.com/ds300/patch-package

Issue #551, opened 2021-01-25
Breaking change happened on v8.3.0 (IE 11 support lost)


jasonwilliams:
This package claims to support IE 11 on the readme. But that stopped working on a minor release.
From what I can gather, v8.3.0 introduced [`stringify.js`](https://github.com/uuidjs/uuid/blob/a91f78d9e593b8a9af1cfebdf33b771b82c42475/src/stringify.js#L13) which utilises default parameters. IE 11 does not support default parameters and thus fails.

I've managed to fix this issue by using babel, but its worth:
- Not using modern syntax in minor releases until checking they work in all browsers you're supporting
- Either remove IE 11 support or remove usage of default params
- Using some rollup plugin to compile down to a compatible version

Related to:
https://github.com/uuidjs/uuid/issues/469

ctavan:
@jasonwilliams thank you for reporting. IE11 is still officially supported.

It sounds like you might be running into the same issue as https://github.com/uuidjs/uuid/issues/516 ? It looks like your bundler is picking up the Node.js build instead of the browser build. Let me know if the suggestions in the other issue work for you as well.

jasonwilliams:
my mainfields is set to `config.resolve.mainFields = ['browser', 'main', 'module'];` which i believe is correct

ctavan:
Well, this explains what you're seeing at least, webpack is picking up the Node.js build: https://github.com/uuidjs/uuid/issues/516#issuecomment-690080241

I believe `'main'` and `'module'` should be swapped as you almost always want to make use of the benefits of ESModules when bundling for the browser (most importantly: tree shaking).

jasonwilliams:
> I believe 'main' and 'module' should be swapped as you almost always want to make use of the benefits of ESModules when bundling for the browser (most importantly: tree shaking).

That was intentional due to an unrelated issue to this.
As long as "browser" is first why would those two being swapped matter?

ctavan:
In the `browser` config we can only specify file overrides but not entry points. In order to support all the various platforms that we want to support, the different fields in `package.json` are crafted the way the are:

* `pkg.main` is picked up when using the module directly in older Node.js versions. This is the Node.js CommonJS build.
* `pkg.module` was originally introduced by webpack for being able to ship an ESM build for bundling. In order to allow webpack builds for Node.js, this points to the Node.js ESModule build.
[...]

Issue #516, opened 2020-09-09
IE 11 breaking with Version 8.3.0


KevinGruber:
## Describe the bug

Since the minor version bump to 8.3.0 we experience problem sin IE11 with uuid.
Based on the error we get it pinpoints to the code which introduced there.
https://github.com/uuidjs/uuid/pull/479/files#diff-016d740d7835ede8d0af47fe304f7191R16
**SCRIPT1006: ')' erwartet** (expected)

## How to reproduce

Use uuid with IE11.

## Expected behavior

It doesn't throw an exception and stops code execution.

## Runtime

- OS: All
- Runtime: IE11

## Additional information

Works fine on other browsers.
We currently solved it with downgrading to 8.2.0.
But it took us quite some time to find the problem to be in uuid, as it was just a minor version bump.

Thanks for looking into it.

ctavan:
@KevinGruber this comes unexpected since the [browser tests for IE11 did pass for the 8.3.0 release](https://github.com/uuidjs/uuid/runs/915976648).

How exactly are you using uuid? Can you provide a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) for the error, including the respective module bundlers etc.? This would help a lot in further debugging the issue.

KevinGruber:
Hi @ctavan,

Of course we are using webpack 4. 
In the bundled code it complains about an closing )

And i found it now with a really small repo.
You are using default value in stringify for parameter offset and that is not allowed in ES5 and lower (IE11).

[IE11 repo.zip](https://github.com/uuidjs/uuid/files/5198113/IE11.repo.zip)

Should I open a PR?
I could do it tomorrow

ctavan:
Hi @KevinGruber. Thanks for providing further information.

Unfortunately to me this still does not explain the issue. Normally, webpack should pick up the files from `dist/esm-browser/` for browser builds, and these files have been transpiled to ES5 in order to work with IE11. Specifically, if you look at https://unpkg.com/browse/uuid@8.3.0/dist/esm-browser/stringify.js line 14 you can see that the default parameter statement has been converted into something that IE11 understands.

So my suspicion is that something about the webpack config might be odd. You could try the examples from https://github.com/uuidjs/uuid/tree/master/examples/browser-webpack (just run `npm start` and open the `example.html` file with IE11) and compare with your setup or post a complete example which involves your exact build config here.

KevinGruber:
Hi @ctavan,

[...]

Issue #314, opened 2019-05-15
Looking for browser object


westlakem:
I'm trying to use this in a nativescript project, and on startup I'm getting the following error: `Cannot read property 'browser' of undefined`

broofa:
The only place this project references `browser` is in the test scripts.   Issue is not in this module.

jbouloux1:
I'm getting the same issue with this module.

`TypeError: Cannot read property 'browser' of undefined
JS:     at Object.../node_modules/pbkdf2/lib/default-encoding.js (file: node_modules/pbkdf2/lib/default-encoding.js:3:11)
JS:     at __webpack_require__ (file: src/webpack/bootstrap:816:0)
JS:     at fn (file: src/webpack/bootstrap:120:0)
JS:     at Object.../node_modules/pbkdf2/index.js (file: node_modules/pbkdf2/index.js:4:22)
JS:     at __webpack_require__ (file: src/webpack/bootstrap:816:0)
JS:     at fn (file: src/webpack/bootstrap:120:0)
JS:     at Object.../node_modules/crypto-browserify/index.js (file: node_modules/crypto-browserify/index.js:14:8)
JS:     at __webpack_require__ (file: src/webpack/bootstrap:816:0)
JS:     at fn (file:///data/data/org.nativescript.D3...`

broofa:
@jbouloux1
1.  Edited your comment to format stacktrace so it's readable
2. Stack trace does not mention `uuid` anywhere.  I.e. this ain't our problem.

In the future, please take a bit more time to verify that the first hit Google returns when you search for an error message is actually relevant before commenting.

ctavan:
@jbouloux1 while I 💯% agree with what @broofa said…

This looks like a problem with your [webpack `resolve.mainFields` config](https://webpack.js.org/configuration/resolve/#resolvemainfields): Apparently you are bundling for the browser but webpack seems to be using the Node.js variant of this library which causes it to pull in a Node.js crypto polyfill. Please check your webpack config.

PR #616, opened 2022-02-10
fix: add `browser` condition to package.json

SimenB:
`jest@28` will ship with `exports` support. But when using `jest-environment-jsdom` to simulate a browser environment, we don't provide the `node` condition, meaning it resolves to `default`. Unless ESM is enabled, this results in a syntax error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,8 +27,7 @@
       },
       "browser": {
         "import": "./dist/esm-browser/index.js",
-        "require": "./dist/index.js",
-        "default": "./dist/esm-browser/index.js"
+        "require": "./dist/index.js"
       },
       "default": "./dist/esm-browser/index.js"
     },
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
         "require": "./dist/index.js",
         "import": "./wrapper.mjs"
       },
-      "browser": {,
+      "browser": {
         "import": "./dist/esm-browser/index.js",
         "require": "./dist/index.js",
         "default": "./dist/esm-browser/index.js"
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
         "require": "./dist/index.js",
         "import": "./wrapper.mjs"
       },
-      "browser": {
+      "browser": {,
+        "import": "./dist/esm-browser/index.js",
         "require": "./dist/index.js",
         "default": "./dist/esm-browser/index.js"
       },
--- a/package.json
+++ b/package.json
@@ -25,6 +25,10 @@
         "require": "./dist/index.js",
         "import": "./wrapper.mjs"
       },
+      "browser": {
+        "require": "./dist/index.js",
+        "default": "./dist/esm-browser/index.js"
+      },
       "default": "./dist/esm-browser/index.js"
     },
     "./package.json": "./package.json"


package.json
```diff
@@ -25,6 +25,11 @@
         "require": "./dist/index.js",
         "import": "./wrapper.mjs"
       },
+      "browser": {
+        "import": "./dist/esm-browser/index.js",
+        "require": "./dist/index.js",
+        "default": "./dist/esm-browser/index.js"
```
ctavan, line 31: (Why) is this default condition needed?

ctavan (commented):
Thank you for raising this!

I _think_ the patch looks correct and non-breaking but I'm still trying to fully understand the consequences, in particular if this will change anything about how webpack and rollup will behave.

A good first sign that the webpack and rollup behaviors remain unchanged is that the [browser tests](https://github.com/uuidjs/uuid/runs/5153076589?check_suite_focus=true) still pass with your patch.

I believe those tests cover the path where webpack and rollup are instructed to use the `esm-browser` build.

Where I'm not entirely sure is cases, where webpack/rollup would pick the `pkg.exports.browser.require` path. I think those cases, the overrides in `pkg.browser` must be taken into account by the bundlers in order to generate a browser-compatible build.

Have you tested this case?

package.json
SimenB: Just to have a fallback if neither of `require` or `import` is provided. Reading https://nodejs.org/api/packages.html#nested-conditions that's not needed tho

SimenB:
> Where I'm not entirely sure is cases, where webpack/rollup would pick the `pkg.exports.browser.require` path. I think those cases, the overrides in `pkg.browser` must be taken into account by the bundlers in order to generate a browser-compatible build.
> 
> Have you tested this case?

Nope, not tested. However, I don't think those fields are related? At least with `main`, it is 100% ignored if `exports` is present, I'd assume `browser` was the same. So older tools without `exports` support use `browser` newer ignores it. Might be wrong, tho 😀

ctavan:
I just ran a quick test: https://github.com/uuidjs/uuid/pull/617/commits/13ceb7949851023efb7ade89caa4928fd6be2f4e

When using `require()` webpack picks `./dist/index.js` as an entry point and then applies the overrides from `pkg.browser`, i.e. it uses `./dist/rng-browser.js` etc.

I have not verified with rollup yet. But at least for webpack, the `pkg.browser` field is _not_ ignored, however after all it does the _right_ thing.

I think this should be covered in the browser tests. I can't promise when I can find the time to do this though.

SimenB:
> I have not verified with rollup yet. But at least for webpack, the `pkg.browser` field is _not_ ignored, however after all it does the _right_ thing.

Their docs are wrong, then? Seems weird 😅 But if it works like you want it to, all is well?

broofa:
> Ah, seems like `browser` is just "aliases", it has nothing directly to do with export conditions

Is it even valid to have a "browser" field nested w/in the "exports" block like this?  If so, where might we find some sort of description or spec for how it's interpreted?

SimenB:
https://nodejs.org/api/packages.html#nested-conditions

ctavan:
> > Ah, seems like `browser` is just "aliases", it has nothing directly to do with export conditions
> 
> Is it even valid to have a "browser" field nested w/in the "exports" block like this? If so, where might we find some sort of description or spec for how it's interpreted?

As @SimenB mentioned the `pkg.browser` field is different from the `browser` environment in `pkg.exports`. Indeed, `pkg.browser` was introduced by browserify a loooong time ago and we've been using it for ages here as well. However, `pkg.exports` also allows a `browser` environment which, according to https://webpack.js.org/guides/package-exports/#target-environment seems to be rather widely supported.

However what I don't really like about the current approach in this PR is that we're mixing to concepts: Through `pkg.exports.browser.require` we're telling bundlers to use the Node.js CommonJS build at `./dist/index.js` and then we're relying on the independent, legacy concept of `pkg.browser` for the browser specific overrides over the Node.js CommonJS build.

I think a cleaner solution would be to provide a consistent CommonJS build for the browser, e.g. in `./dist/commonjs-browser/index.js`. Folks have run into similar issues in the past ([1](https://github.com/uuidjs/uuid/issues/516#issuecomment-690080241), [2](https://github.com/uuidjs/uuid/issues/551#issuecomment-766857268), [3](https://github.com/uuidjs/uuid/issues/314#issuecomment-737761009), [4](https://github.com/uuidjs/uuid/issues/314#issuecomment-737761009)) when misconfiguring their bundlers… I would still argue that this is _always_ an unintended misconfiguration and should not be done, folks should not choose the CommonJS builds of their dependencies for browser builds, they should use the ESM build that natively supports tree shaking instead.

However in this case now with `jest-environment-jsdom` where a browser env is only simulated within a Node.js process things may actually start looking differently. Thinking about this, however, uncovers an interesting aspect: I believe with this current patch and when using `jest-environment-jsdom` jest may actually end up executing the CommonJS Node.js code and not the browser code. In practice, this will probably produce the intended results for the test cases. However it's still possible that such test cases will pass with Jest but would fail in a browser since browsers will effectively execute different code.

I'm sorry for the long write up, I guess the main message is that this is _complicated_…

@SimenB do you have an example of where a syntax error is thrown (what you reference in the original PR description)? I would really like to figure out what happens under the hood when jest runs these tests.

SimenB:
A browser build in CJS would be even better of course 🙂 

>@SimenB do you have an example of where a syntax error is thrown (what you reference in the original PR description)? I would really like to figure out what happens under the hood when jest runs these tests.

Jest provides `['require', 'browser', 'default']` as conditions, meaning `default` is returned, which is `./dist/esm-browser/index.js`.

https://github.com/SimenB/jest-uuid

Remove comment at top of `test.js` to see it pass (as Jest then runs in node env and provides `node` condition instead of `browser`)

SimenB:
@ctavan how do we unblock this? 🙂 This issue is actually stopping me from using the Jest alpha at work 😅 (although I guess I can just patch `package.json`, but still)

ctavan:
Sorry for the silence and sorry to hear this blocks you.

I'm happy to help unblock this but I have to admit that I'm not sure what the right way forward is. I experimented with building a browser commonjs build, but that only reveals the actual problem…

If you run jest on `jsdom.test.js` in https://github.com/uuidjs/uuid/tree/simenb-patch/examples/jest-uuid-main you get:

```
$ ./node_modules/.bin/jest jsdom.test.js
  console.log
    commonjs-browser RNG

      at Object.<anonymous> (../../dist/commonjs-browser/rng.js:3:9)

 FAIL  ./jsdom.test.js
  ● Test suite failed to run

    crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported

      20 |
      21 |     if (!getRandomValues) {
    > 22 |       throw new Error('crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported');
         |             ^
      23 |     }
      24 |   }
      25 |

      at rng (../../dist/commonjs-browser/rng.js:22:13)
      at Object.v4 (../../dist/commonjs-browser/v4.js:23:32)
      at Object.<anonymous> (jsdom.test.js:5:18)
```

So the browser commonjs build uses browser APIs (`crypto`), it has to do this, otherwise we can't sell that as a browser build. However those APIs don't seem to be provided/mocked by jsdom.

I'm currently failing to understand what the expectations are when you run a test case with `@jest-environment jsdom`? How close to a browser environment is this supposed to be?

If it's not reasonably close, I think that resolving the `pkg.exports.browser` target for `@jest-environment jsdom` tests will inevitably cause many problems like the one above, where the code that is offered on the `pkg.exports.browser` export relies on native browser APIs.

Can you help clarify the expectations?

SimenB:
> So the browser commonjs build uses browser APIs (`crypto`), it has to do this, otherwise we can't sell that as a browser build. However those APIs don't seem to be provided/mocked by jsdom.

That seems perfectly fine to me? "If you want to use `uuid` in an "old" browser, you need to polyfill". I don't see the issue with that.

> I'm currently failing to understand what the expectations are when you run a test case with `@jest-environment jsdom`? How close to a browser environment is this supposed to be?

As close as possible. If APIs are missing from the env, the module should complain. The fact I want to use CJS instead of ESM is separate from browser vs node. It's not up to this module to say "you cannot use `require` in the browse", it should just say "here's a browser entry point for CJS" and it's up to the consumer to ensure their environment works for what they're trying to load.

ro-savage:
Jest 28 has now been released - which means that Jest and UUID are no longer compatible.

For anyone trying to fix this issue - @dbjorge did a good temp solution here: https://github.com/microsoft/accessibility-insights-web/pull/5421#issuecomment-1109168149

But would be great to have a proper solution to this.

ctavan:
Looping in @dbjorge here since I think this thread is more suitable to discuss the jest/uuid specific issue than https://github.com/uuidjs/uuid/pull/616#issuecomment-1043492062.

First of all, thanks for the great summary @dbjorge!

At the risk of repeating what @dbjorge already summarized perfectly well, I'd like to highlight again:

Jest + UUID just "accidentally" worked in the past because Jest used to pick the Node.js build even in environments like jsdom that were supposed to mimic the browser. Apparently this has worked well for folks in practice, so I think a [custom jest resolver like this one](https://github.com/microsoft/accessibility-insights-web/pull/5421/commits/9ad4e618019298d82732d49d00aafb846fb6bac7) is actually a great mitigation for the time being as it just puts us back to where we were before Jest 28.

If we were adding a CommonJS browser build for completeness of the support matrix of `uuid`, we would still need something like https://github.com/jsdom/jsdom/pull/3352 to make this work with Jest 28. And I stand with my statement that I generally think CommonJS for the browser has no production use case. Requiring a whole new build specifically for testing purposes which has otherwise no production use case to me sounds more like something that should be fixed in the testing library rather than in the libraries being tested… So to that end I'd prefer see ESM support land in Jest.

@SimenB two questions:

1. Can you foresee if/when Jest will get ESM support?
2. Assuming, 1. will materialize eventually, could Jest potentially ship fallback resolutions logic for widely spread libraries like `uuid` that would fall back to resolving the node build (essentially bundling the [custom resolver that @dbjorge wrote]((https://github.com/microsoft/accessibility-insights-web/pull/5421/commits/9ad4e618019298d82732d49d00aafb846fb6bac7))?)?

I can see how this affects many people, given both jest and uuid are so widespread, I'm just still failing to clearly identify the _right_ solution for this situation.

dbjorge:
@ctavan Thanks for the ping!

https://github.com/facebook/jest/issues/9430 is the meta-issue tracking ESM support in Jest. It is a very long and very [soul-crushing](https://github.com/facebook/jest/pull/11529#issuecomment-1027091448) read. For the purposes of our discussion here, I think it's probably safe to say that it won't happen soon enough for the folks that need Jest+uuid workarounds to unblock Jest 28 updates.

I want to emphasize that, as a user of both libraries, I don't think anyone on the uuid _or_ Jest sides are making unreasonable decisions here; I don't really feel convinced that a workaround really "belongs" in either library. Particularly:

* I think it's reasonable for the `uuid` maintainers to push back on adding a CommonJS + browser export just for the sake of one testing framework
* I think the new Jest behavior to try to use browser exports from `jest-environment-jsdom` is an excellent change overall
* If I were a uuid maintainer, I would be very hesitant to touch its exports unnecessarily for fear of breaking backcompat with someone's weird webpack config
* If I were a Jest maintainer, I would be very hesitant to establish a precedent of adding package-specific magic to Jest's default resolver

I think there are probably a large number of users that will be impacted by this, but more importantly than that, I want to recognize that neither Jest nor uuid have any obligation to be cross-compatible with one another. I am a thankful user of the work you have both made available for free, and neither project's maintainers owe me (and certainly not my employer) a damn thing; if I have to maintain a few lines of `resolver.js` to use them together, that's perfectly fine.

ctavan:
Thanks for chiming in @SimenB!

> I don't think we'll be adding workarounds for any libraries directly in Jest. The easiest solution would probably be for somebody to create a `jest-uuid` that people can stick in the [`moduleNameMapper`](https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring) config. It could just do `module.exports = require('uuid')` and possibly be a cleaner solution than custom resolvers (as it composes better if people already use a custom resolver).

@dbjorge what do you think about this approach? Would this be something you'd consider to test-drive on [accessibility-insights-web](https://github.com/microsoft/accessibility-insights-web)?

> Or potentially having some well maintained custom resolver people can plug in that looks like the one @dbjorge has built, but supporting more modules than just `uuid`. 

Or maybe rather this second approach?

Given I currently don't have my own use case for this, I don't think I'll be able to provide a solution (like the `jest-uuid` wrapper), but I'd be happy to contribute (e.g. I'd be open to host such a module in the `uuidjs` github & npm org).

dairyisscary:
For anyone coming here looking for a quick workaround, i did the following in my jest config (which i think is more or less @SimenB's suggestion with a  `jest-uuid`)

```js
moduleNameMapper: {
  "^uuid$": "<rootDir>/node_modules/uuid/dist/index.js",
},
```

dbjorge:
> > The easiest solution would probably be for somebody to create a `jest-uuid` that people can stick in the [`moduleNameMapper`](https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring) config. It could just do `module.exports = require('uuid')` and possibly be a cleaner solution than custom resolvers (as it composes better if people already use a custom resolver).
> 
> @dbjorge what do you think about this approach? Would this be something you'd consider to test-drive on [accessibility-insights-web](https://github.com/microsoft/accessibility-insights-web)?

I actually tried basically this solution before I went with the custom resolver - I don't think a wrapper package is necessary, you can just do something like:

```js
moduleNameMapper: {
  // This forces Jest/jest-environment-jsdom to use a Node+CommonJS version of uuid, not a Browser+ESM one
  // See https://github.com/uuidjs/uuid/pull/616
  '^uuid$': require.resolve('uuid'),
}
```

However, I rejected this in favor of the custom resolver because my understanding of the `moduleNameMapper` option is that it would essentially force every transitive dependency on uuid in my project to resolve to whichever version of uuid happens to be hoisted to the root `node_modules` folder. I'm already working in a project which has multiple versions of uuid in its dependency tree, but even if I wasn't, this scenario creates a maintenance time bomb if `uuid` ever releases a v9 with breaking changes. I preferred the custom resolver because it lets each dependency chain continue to use (the CommonJS version of) whatever `uuid` it was already using.

SimenB:
@dbjorge your analysis is correct though, this forces a single version across the entire tree

zak39:
> For anyone coming here looking for a quick workaround, i did the following in my jest config (which i think is more or less @SimenB's suggestion with a `jest-uuid`)
> 
> ```js
> moduleNameMapper: {
>   "^uuid$": "<rootDir>/node_modules/uuid/dist/index.js",
> },
> ```

Thanks @dairyisscary :)
I duplicated by mistake this issue : https://github.com/facebook/jest/issues/12790

I added your suggestion and it works ! :muscle: 
But, Do developer of Jest will fix it ? Or the `uuid` module is useless ?

zak39:
> @zak39 The answer is summarized here [#616 (comment)](https://github.com/uuidjs/uuid/pull/616#issuecomment-1110284432) - please make sure to skim the discussion first.
> 
> And also, let's be considerate and not calling the library `useless`.

Yeah, I'm sorry, I would not mean this library is "useless". Indeed, the word is stronger... I would mean this library could be replaced or something like this.

And thanks, I didn't see this comment. I skimed it ^^'

SimenB:
> Sorry for the silence and sorry to hear this blocks you.
> 
> I'm happy to help unblock this but I have to admit that I'm not sure what the right way forward is. I experimented with building a browser commonjs build, but that only reveals the actual problem…
> 
> If you run jest on `jsdom.test.js` in https://github.com/uuidjs/uuid/tree/simenb-patch/examples/jest-uuid-main you get:
> 
> ```
> $ ./node_modules/.bin/jest jsdom.test.js
>   console.log
>     commonjs-browser RNG
> 
>       at Object.<anonymous> (../../dist/commonjs-browser/rng.js:3:9)
> 
>  FAIL  ./jsdom.test.js
>   ● Test suite failed to run
> 
>     crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported
> 
>       20 |
>       21 |     if (!getRandomValues) {
>     > 22 |       throw new Error('crypto.getRandomValues() not supported. See https://github.com/uuidjs/uuid#getrandomvalues-not-supported');
>          |             ^
>       23 |     }
>       24 |   }
>       25 |
> 
>       at rng (../../dist/commonjs-browser/rng.js:22:13)
>       at Object.v4 (../../dist/commonjs-browser/v4.js:23:32)
>       at Object.<anonymous> (jsdom.test.js:5:18)
> ```
> 
> So the browser commonjs build uses browser APIs (`crypto`), it has to do this, otherwise we can't sell that as a browser build. However those APIs don't seem to be provided/mocked by jsdom.
> 
> I'm currently failing to understand what the expectations are when you run a test case with `@jest-environment jsdom`? How close to a browser environment is this supposed to be?

JSDOM v20 supports `crypto.getRandomValues()`, fwiw

broofa:
> JSDOM v20 supports crypto.getRandomValues(), fwiw

@SimenB I'm being lazy here (headed out the door to walk dog, get my day started), but does that mean things should "just work" now?  ... or that work is still needed w/in this module to resolve this issue?

ctavan:
I can confirm that this now works with `jest-environment-jsdom@29.0.0-alpha.0` (which has ` jsdom@20.0.0` as a dependency).

ctavan:
I have released `uuid@9.0.0-beta.0` that contains this patch. Please try it out and let me know if it fixes the Jest interoperability (to be tested with `jest@29.0.0-alpha.1`).

monsieurnebo:
> I have released `uuid@9.0.0-beta.0` that contains this patch. Please try it out and let me know if it fixes the Jest interoperability (to be tested with `jest@29.0.0-alpha.1`).

I just gave it a try and it seems to fix the issue on my side. Looking forward for the stable release(s).

Thanks for your work 💪

SimenB:
@ctavan can confirm it allows us to remove the `moduleMapper` config at work as well 👍

SimenB:
FWIW, Jest 29 is released (which means `jest-environment-jsdom@29` is out, which comes with `jsdom@20`)

anandhsonu1994:
@SimenB @ctavan, When can we expect the release?

SimenB:
I have no idea, Jest 29 has been out for a week

monsieurnebo:
> FWIW, Jest 29 is released (which means `jest-environment-jsdom@29` is out, which comes with `jsdom@20`)

Thanks for the info. Awaiting UUID v9 now then :)

PR #618, opened 2022-03-08
fix: unhandle error when parameter is not set in v3 and v5

rzkytmgr:
[source](https://github.com/uuidjs/uuid/blob/main/src/v35.js#L25) only checks if `namespace` type is equal to `string`. The error will be triggered when the `namespace` type is not equal to `string` unless the `namespace` parameter presented as an array.

Thank you,

### DIFF
--- a/test/unit/v35.test.js
+++ b/test/unit/v35.test.js
@@ -148,6 +148,24 @@ describe('v35', () => {
     assert.deepEqual(buf, ['landmaster', 'landmaster', 'landmaster'].concat(testBuf));
   });
 
+  test('v3 undefined/null', () => {
+    assert.throws(() => {
+      v3();
+    });
+
+    assert.throws(() => {
+      v3('hello');
+    });
+
+    assert.throws(() => {
+      v3('hello.example.com', undefined);
+    });
+
+    assert.throws(() => {
+      v3('hello.example.com', null, new Array(16));
+    });
+  });
+
   test('v5', () => {
     // Expect to get the same results as http://tools.adjet.org/uuid-v5
     assert.strictEqual(v5('hello.example.com', v5.DNS), 'fdda765f-fc57-5604-a269-52a7df8164ec');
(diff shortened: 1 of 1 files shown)


broofa (changes requested):
Can you add a unit test [here](https://github.com/uuidjs/uuid/blob/main/test/unit/v35.test.js), to test the undefined / null `namespace` cases?

src/v35.js
```diff
@@ -26,7 +26,7 @@ export default function v35(name, version, hashfunc) {
       namespace = parse(namespace);
     }
 
-    if (namespace.length !== 16) {
+    if (namespace?.length !== 16) {
       throw TypeError('Namespace must be array-like (16 iterable integer values, 0-255)');
     }
```
broofa, line 31:
```suggestion
if (!Array.isArray(namespace) || namespace.length !== 16) {
      throw TypeError('namespace must be uuid string or an Array-like of 16 byte values');
    }```

rzkytmgr:
> Can you add a unit test [here](https://github.com/uuidjs/uuid/blob/main/test/unit/v35.test.js), to test the undefined / null `namespace` cases?

of course with pleasure, let me work for it.

src/v35.js
rzkytmgr: When I applied the suggested change you've made, there are total 8 failed test cases in `test/unit/v35.test.js`.
There are several test cases where the namespace parameter is given a string argument, while the above code checks if it is not an array then a TypeError will be throwed.

broofa: Gah, sorry... I forgot `Array.isArray()` doesn't recognize typed arrays.  ("Wat???")

So, yeah, your original change (`?.`) should be fine.  Sorry for the confusion.

broofa (approved):

broofa:
@ctavan, You okay merging this in this form (with CI tweaked to run on 16.13)?   See #619 for details.  tl;dr: It's an issue in our example code, not something @rzkytmgr introduced.

broofa:
@ctavan *ping*

ctavan:
Proposed an alternative fix to the CI issue in https://github.com/uuidjs/uuid/pull/621.

I'd prefer to merge that one first to produce a cleaner diff in this PR.

ctavan:
I think I somehow messed this up while rebasing… I created #622 with the same patch instead.

PR #622, opened 2022-03-17
fix: handle error when parameter is not set in v3 and v5

ctavan:


LinusU (approved):
👍

broofa:
@rzkytmgr Thanks for the help!

PR #623, opened 2022-03-17
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/src/stringify.js
+++ b/src/stringify.js
@@ -7,7 +7,7 @@ import validate from './validate.js';
 const byteToHex = [];
 
 for (let i = 0; i < 256; ++i) {
-  byteToHex.push((i + 0x100).toString(16).substr(1));
+  byteToHex.push((i + 0x100).toString(16).slice(1));
 }
 
 export function unsafeStringify(arr, offset = 0) {
--- a/src/version.js
+++ b/src/version.js
@@ -5,7 +5,7 @@ function version(uuid) {
     throw TypeError('Invalid UUID');
   }
 
-  return parseInt(uuid.substr(14, 1), 16);
+  return parseInt(uuid.slice(14, 15), 16);
 }
 
 export default version;


LinusU (approved):
And the bundle gets one byte smaller hehe 🎉

broofa:
@ctavan: Just a heads up that I removed the `bundlewatch` check, as that's been stalling on recent PRs because... well... we don't seem to actually have that as part of any actions. 🤷 
CleanShot 2022-03-17 at 08 19 40@2x

And while I was in there, I added `ci (16.x)` as a required check:
CleanShot 2022-03-17 at 08 21 38@2x
 
Let me know if I should not have done these things for any reason.

ctavan:
Bundlewatch only works for branches on this repo since it needs an api key that could otherwise be exfiltrated when run on branches of forks. The same is true for browserstack.

Would be great if you could bring bundlewatch back!

Another nit: since we try to generate change logs and derive version jumps from the commit messages, it would be great to always make sure to follow the https://www.conventionalcommits.org/en/v1.0.0/ format when squash merging branches.

broofa:
Ah, okay... I was confused where that `bundlewatch` check was coming from.  Didn't realize we had it set up as an [external service](https://github.com/bundlewatch/service).

I've restored the checks back to their prior state.

> always make sure to follow the Conventional Commits format

Yeah, my bad.  I noticed that right after mashing the Merge button. 😢

ctavan:
Yeah, it’s a pity that our precommit linting checks, which enforce well formed commit messages locally, don’t apply when merging…

PR #624, opened 2022-03-25
Delete empty file: a

jonpulsifer:
:wave: found a nice drive-by!

a: This file looks to have been accidentally committed, is blank, and should be 🔥

Ref https://github.com/uuidjs/uuid/pull/621


ctavan:
Thank you for the cleanup!

PR #627, opened 2022-04-29
Update Readme to cover Expo SDK45+

hirbod:
I just updated the README to cover another method how to use this package with Expo SDK 45+ and without random-helper.

Since expo-random ported to JSI, this should be faster in theory.


broofa:
As noted at [the top of README](https://github.com/uuidjs/uuid/blob/main/README.md?plain=1#L1-L3), edits need to be made to README_js.md.  Then `npm run docs`.

That said, the Known Issues section of the README should probably be refactored into a separate file.  Or maybe a wiki page?  @ctavan

hirbod:
Sorry, I'm not going to setup the whole tooling for a README file.

broofa:
No worries.  Thanks for taking the time to let us know about the issue.  I'll revisit this if/when we refactor our README.

PR #637, opened 2022-05-23
chore: use github form syntax for better issue templates

broofa:
'Noticed that `npm/cli` is using yaml-based templates for their issues, which make for a significantly better user-experience.  Most importantly, hopefully this helps get the message across to users that they need to provide actual, reproducible recipes instead of merely copy/pasting their stack traces.

Here's what the bug report will look like (more or less): https://github.com/uuidjs/uuid/blob/67d48bc2d613bf8129dd3785b64579715e794aaa/.github/ISSUE_TEMPLATE/bug_report.yml


ctavan (approved):
Nice! I wasn't aware of this feature yet!

Issue #614, opened 2022-02-05
TypeError: undefined is not an object (evaluating 'rnds[6]')


MennatAllahMagdy:
## Describe the bug

When I use v4 from uuid, I got that error 'TypeError: undefined is not an object (evaluating 'rnds[6]')'

## How to reproduce

import { v4 as uuidv4 } from 'uuid';
const Uid = uuidv4();

## Runtime

- OS: [WSL2]
- "react-native": "0.64.3"
- "uuid": "^8.3.2"
- "@types/uuid": "^8.3.2"

nianxi:
Same issue with you.
Working online but collects errors everyday:
TypeError: Cannot read property '6' of undefined

broofa:
@nianxi : Can you provide complete instructions for how to reproduce this?

stefreak:
For me, this works when "Debug with Chrome" in my expo app is enabled, and when "Debug with Chrome" is disabled I get the error. Using expo and running the app in iOS Simulator

stefreak:
Update: For me, the issue seems to be gone after installing `react-native-get-random-values`.

PR #641, opened 2022-07-13
Fix undefined return from crypto.getRandomValues in older Safari

LBRDan:
In some envs (like Safari 6), `crypto.getRandomValues` does not return anything and as per spec it changes its argument in place
This behavior it's specified (not clearly) in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) right at the beginning of the paragraph:
`The array given as the parameter is filled with random numbers`

The current implementation is right on the current web crypto [specs](https://w3c.github.io/webcrypto/#Crypto-method-getRandomValues) (point 4).

This version should work across all window.crypto browsers and IE (`msCrypto`)

### DIFF
--- a/src/rng-browser.js
+++ b/src/rng-browser.js
@@ -24,6 +24,8 @@ export default function rng() {
       );
     }
   }
+  
+  getRandomValues(rnds8)
 
-  return getRandomValues(rnds8);
+  return rnds8;
 }


LBRDan:
It could affect and fix even this #614

broofa:
Thanks for the contribution, however Safari 6 is not a supported browser (last release was Aug 2015, Apple hasn't supported it since MacOS 10.8, circa 2015-2016).

Re: 614, it's not clear that's the same problem.  Users there are on react-native and/or Chrome so... 🤷.

I know this is a trivial change to make, however this is a stable project with a very large number of users, and I'd rather not set the precedent of churning code for unsupported platforms.

PR #642, opened 2022-07-29
feat: add Jest/jsdom compatibility

### LINKED ISSUE #451: ESM Problems with Jest@26 and Node.js 14.x

ctavan:
BREAKING CHANGE: Jest runs browser tests in a jsdom environment which now also supports web crypto polyfills.

Since ESM support in Jest is currently still limited, it requires a commonJS build for browser environments, see the discussion in https://github.com/uuidjs/uuid/pull/616 for all the details.

Closes #617
Closes #616
Fixes #451

Co-authored-by: Christoph Tavan <dev@tavan.de>


SimenB:
yay, thanks!

scripts/build.sh
```diff
@@ -11,8 +11,12 @@ DIR="$ROOT/dist"
 rm -rf "$DIR"
 mkdir -p "$DIR"
 
-# Transpile CommonJS versions of files
-babel --env-name commonjs src --source-root src --out-dir "$DIR" --copy-files --quiet
+# Transpile CommonJS versions of files for node
+babel --env-name commonjsNode src --source-root src --out-dir "$DIR" --copy-files --quiet
+
+# Transpile CommonJS versions of files for node
```
SimenB, line 17:
```suggestion
# Transpile CommonJS versions of files for the browser
```

.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x, 16.x]
+        node-version: [12.x, 14.x, 16.x, 18.x]
```
SimenB, line 12: Jest 29 will drop 12, fwiw. Might affect integration tests?

ctavan: I think in here we're not yet ready to drop 12.x support. Do you know that this will break?

scripts/build.sh
```diff
@@ -21,11 +25,19 @@ babel --env-name esmBrowser src --source-root src --out-dir "$DIR/esm-browser" -
...
 rm -rf "$DIR/esm-browser/uuid-bin.js"
 rm -rf "$DIR/esm-node/bin"
 rm -rf "$DIR/esm-node/uuid-bin.js"
 
+for FILE in "$DIR"/commonjs-browser/*-browser.js
+do
+    echo "Replacing node-specific file for commonjs-browser: $FILE"
+    mv "$FILE" "${FILE/-browser.js/.js}"
+done
+
```
broofa, line 40: Can you add a comment here explaining *why* we do this?

broofa (approved):

.github/workflows/ci.yml
SimenB: We've stopped testing v12, and setup babel to target v14. Not any explicit breaking changes, tho.

JSDOM v20 also has minimum node@14, fwiw

scripts/build.sh
ctavan: Added a longer description of what we're doing here.

ctavan:
While I don't have proof of it, I think this change to the `pkg.exports` field may in fact be a breaking change for some. I'm therefore marking this as breaking change and we'll only get this change with the v9 release.

SimenB:
Node version minimum makes this a breaking change, imo 😀

However, if you ignore that, I don't think this is breaking. People requesting `browser` _and_ `require` will now get what they want instead of ESM. If they don't want cjs, it should be requested. And people requesting all of `browser`, `import` and `require` will still get the ESM version as it's above the `require` condition. So the `exports` changes in isolation are not, IMO, breaking.

broofa:
Let's push this out in v9, and get v9 rolled out ASAP.  I've cleaned up the [project board for it](https://github.com/uuidjs/uuid/projects/1), so there's only one other pending pr (#604).  @ctavan: Is that something you'd be able to close out sometime soon?

If so, let's merge both these PRs and get v9 rolled out.

PR #645, opened 2022-08-03
feat: remove UMD build

### LINKED ISSUE #620: npm package contains minified files
## Is your feature request related to a problem? Please describe.

The package as it is published on npm currently contains multiple minified files; this is problematic from a security auditing perspective, as minified code is difficult to audit. This is particularly problematic because of the often security-sensitive nature of the `uuid` module.

A more detailed explanation of the problem with minified builds and why they are unnecessary to publish in the first place, can be found [here](https://gist.github.com/joepie91/04cc8329df231ea3e262dffe3d41f848).

## Describe the solution you'd like

Removing all minified builds from the npm package in a future release, instead either a) providing instructions for users on how to obtain a minified build, or b) if necessary for certain CDNs, publishing the minified builds in a separate package that users wouldn't install into their project. 

## Describe alternatives you've considered

- __Reproducing from source:__ Requires auditing the entire build stack due to potential supply chain attacks, with build tools being able to modify the code going through them.
- __Prettifying minified code and auditing that:__ Cumbersome, unreliable because a lack of meaningful names makes it hard to understand the code correctly.
- __Just not using the minified files:__ For a reliable dependency audit, *all* files present in the dependency tree must be audited, because there is (AFAIK) no runtime-level mechanism to block certain files from being loaded in eg. a multi-stage attack.

## Additional context

N/A

ctavan:
BREAKING CHANGE: Remove the minified UMD build from the package.

Minified code is hard to audit and since this is a widely used library
it seems more appropriate nowadays to optimize for auditability than to
ship a legacy module format that, at best, serves educational purposes
nowadays.

For production browser use cases, users should be using a bundler. For
educational purposes, today's online sandboxes like replit.com offer
convenient ways to load npm modules, so the use case for UMD through
repos like UNPKG or jsDelivr has largely vanished.

Fixes #620


broofa (approved):
CleanShot 2022-08-03 at 16 12 28@2x

CleanShot 2022-08-03 at 16 12 28@2x
CleanShot 2022-08-03 at 16 12 28@2x

image

PR #647, opened 2022-08-11
chore: enable Close Stale Issues action

broofa:
I think/hope this will enable the [Close Stale Issues](https://github.com/marketplace/actions/close-stale-issues) workflow, configured to mark issues as stale after 90 days, and close 30 days later.  Won't know until it's actually on main.  We could do something similar for PRs, but I don't see an immediate need for that.

@ctavan: This would save the occasional (and irregular) sweep of issues I find myself doing manually from time to time.


ctavan (approved):
Thanks a lot. This looks very good!

PR #648, opened 2022-08-12
[DO NOT MERGE] Attempt to see if a github action secret can be read from PR

copinf22:
Depending on how it is configured, this secret https://github.com/uuidjs/uuid/blob/f2826b6bc35bb8df512a9dc4fe5236050b4d3634/.github/workflows/ci.yml#L40 may be able to be exposed by opening a pull request with a malicious package.json. I cannot tell from past PRs from external forks if this is possible or not, because the latest ones with any runs have their logs expired which would tell if bundlewatch failed to authenticate or not because the token wasnt available to it. If it is exposed, i will open an issue, otherwise I will just delete this PR. Sorry if this seems spammy :(


copinf22:
Ah, it seems to be configured so "First-time contributors need a maintainer to approve running workflows" so this will not execute.

PR #653, opened 2022-09-03
chore: drop node 10.x to upgrade dev dependencies

ctavan:
BREAKING CHANGE: This library always aims at supporting one EOLed LTS release
which by this time now is 12.x which has reached EOL 30 Apr 2022.

The actual change was already introduced in #643 but the commit was missing the
breaking change label.


broofa (approved):
🚢

Issue #606, opened 2021-12-07
Have validate accept null and undefined


grosch:
Just to make life easier, it would be great if when I call `uuidValidate` and pass in `null` or `undefined`, it returned false.  Right now the compiler doesn't allow a potentially null/undefined value to be given.

Right now I'm doing something like this:

```typescript
this.route.paramMap
    .pipe(
        map(x => x.get('id')),
        filter(x => !!x && uuidValidate(x)),
        mustHaveValue(),
        switchMap(x => this.clientService.get(x)),
        untilDestroyed(this)
    )
```

ctavan:
I think it could make sense to fix that on

LinusU:
If you are sending a PR, I would recommend to change it to from `(uuid: string) => boolean` to `(uuid: any) => uuid is string`. That way TypeScript will know that you can pass anything in to the `validate` function, and that if `true` was returned it was definitely a `string`.

broofa:
@LinusU Can you provide a link to the typescript docs that discuss whatever that behavior is that you're describing?  I'm still learning TS and this sort of conditional typing sounds interesting.

grosch:
> @LinusU Can you provide a link to the typescript

https://www.typescripttutorial.net/typescript-tutorial/typescript-type-guards/

LinusU:
some more refs:

https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates
https://stackoverflow.com/q/40081332/148072

ctavan:
Reopening as this may be solved in #654.

broofa:
Closing.  As it stands, this is an `@types/uuid` issue.   If/when we integrate TS support here, this should get fixed so I consider this a dupe of #654.

PR #654, opened 2022-09-07
feat: add TypeScript typings

LinusU:
I've made this typings by copying in everything from the readme and then formatting it properly. Opening as a draft for now as I would like to investigate some way to make sure that the typings and readme are always in sync. Potentially [ts-readme-generator](https://github.com/LinusU/ts-readme-generator) or a similar script. Otherwise I fear that the typings and readme file will drift away from each other 😅 

- [x] Write typings
- [ ] Test readme with `ts-readme-generator`
    - [x] Add support for `@throws`
    - [x] Add support for `@example`
    - [x] Add support for numeric literals
    - [x] Add support for `@note`
    - [ ] Investigate why `V1Options`/`V4Options` isn't expanded
    - [ ] Handle overloaded functions
    - [ ] Fix `@example` (and others) on exported constants
- [ ] Test typings with `tsd`
- [ ] Remove `runmd` dependency and `README_js.md` file

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,11 @@
         "import": "./dist/esm-browser/index.js",
         "require": "./dist/commonjs-browser/index.js"
       },
-      "default": "./dist/esm-browser/index.js"
+      "default": "./dist/esm-browser/index.js",
+      "types": {
+        "module": "./types.d.mts",
+        "default": "./types.d.ts"
+      }
     },
     "./package.json": "./package.json"
   },
--- a/types.d.ts
+++ b/types.d.ts
@@ -1,0 +1,241 @@
+/**
+ * The nil UUID string (all zeros)
+ *
+ * @example
+ * ```javascript
+ * import { NIL as NIL_UUID } from 'uuid';
+ *
+ * NIL_UUID; // ⇨ '00000000-0000-0000-0000-000000000000'
+ * ```
+ */
+export const NIL: '00000000-0000-0000-0000-000000000000'
+
+/**
+ * Convert UUID string to array of bytes
+ *
+ * @param {string} str - A valid UUID string
+ * @throws {TypeError} If `str` is not a valid UUID
+ *
+ * @note Ordering of values in the byte arrays used by `parse()` and `stringify()` follows the left ↠ right order of hex-pairs in UUID strings. As shown in the example below.
+ *
+ * @example
+ * ```javascript
+ * import { parse as uuidParse } from 'uuid';
+ *
+ * // Parse a UUID
+ * const bytes = uuidParse('6ec0bd7f-11c0-43da-975e-2a8ad9ebae0b');
+ *
+ * // Convert to hex strings to show byte order (for documentation purposes)
+ * [...bytes].map((v) => v.toString(16).padStart(2, '0')); // ⇨
+ *   // [
+ *   //   '6e', 'c0', 'bd', '7f',
+ *   //   '11', 'c0', '43', 'da',
+ *   //   '97', '5e', '2a', '8a',
+ *   //   'd9', 'eb', 'ae', '0b'
+ *   // ]
+ * ```
+ */
+export function parse(str: string): Uint8Array
+
+/**
+ *
+ * @param arr Collection of 16 values (starting from `offset`) between 0-255
+ * @param offset Starting index in the Array
+ * @throws {TypeError} If a valid UUID string cannot be generated
+ *
+ * @note Ordering of values in the byte arrays used by `parse()` and `stringify()` follows the left ↠ right order of hex-pairs in UUID strings. As shown in the example below.
+ *
+ * @example
+ * ```javascript
+ * import { stringify as uuidStringify } from 'uuid';
+ *
+ * const uuidBytes = [
+ *   0x6e, 0xc0, 0xbd, 0x7f, 0x11, 0xc0, 0x43, 0xda, 0x97, 0x5e, 0x2a, 0x8a, 0xd9, 0xeb, 0xae, 0x0b,
+ * ];
+ *
+ * uuidStringify(uuidBytes); // ⇨ '6ec0bd7f-11c0-43da-975e-2a8ad9ebae0b'
+ * ```
+ */
+export function stringify(arr: ArrayLike<number>, offset?: number): string
+
+interface V1Options {
+  node?: ArrayLike<number>
+  clockseq?: number
+  msecs?: number
+  nsecs?: number
+  random?: ArrayLike<number>
+  rng?: () => ArrayLike<number>
+}
+
+/**
+ * Create an RFC version 1 (timestamp) UUID
+ *
+ * @param options `Object` with one or more of the following properties:
+ * @param options.node RFC "node" field as an Array[6] of byte values (per 4.1.6)
+ * @param options.clockseq RFC "clock sequence" as a `Number` between 0 - 0x3fff
+ * @param options.msecs RFC "timestamp" field (`Number` of milliseconds, unix epoch)
+ * @param options.nsecs RFC "timestamp" field (`Number` of nanseconds to add to `msecs`, should be 0-10,000)
+ * @param options.random `Array` of 16 random bytes (0-255)
+ * @param options.rng Alternative to `options.random`, a `Function` that returns an `Array` of 16 random bytes (0-255)
+ * @param buffer If specified, uuid will be written here in byte-form, starting at `offset`
+ * @param offset Index to start writing UUID bytes in `buffer`
+ * @returns UUID `String` if no `buffer` is specified, otherwise returns `buffer`
+ * @throws {Error} If more than 10M UUIDs/sec are requested
+ *
+ * @note The default [node id](https://tools.ietf.org/html/rfc4122#section-4.1.6) (the last 12 digits in the UUID) is generated once, randomly, on process startup, and then remains unchanged for the duration of the process
+ * @note `options.random` and `options.rng` are only meaningful on the very first call to `v1()`, where they may be passed to initialize the internal `node` and `clockseq` fields
+ *
+ * @example
+ * ```javascript
+ * import { v1 as uuidv1 } from 'uuid';
+ *
+ * uuidv1(); // ⇨ '2c5ea4c0-4067-11e9-8bad-9b1deb4d3b7d'
+ * ```
+ *
+ * @example
+ * ```javascript
+ * import { v1 as uuidv1 } from 'uuid';
+ *
+ * const v1options = {
+ *   node: [0x01, 0x23, 0x45, 0x67, 0x89, 0xab],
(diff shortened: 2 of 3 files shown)


ctavan:
I think this looks really good, although – given my lack of TS experience – I can't judge the quality of the types. Questions that come to my mind:

- Does this solve the issue raised in #606 which seems to exist in the current DefinitelyTyped types?
- In the [current Types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/uuid/index.d.ts) we use rather complicated interfaces. What are the pros and cons of something like that?
- Also big +1 to auto-generating the docs from the types!

package.json
```diff
@@ -29,7 +29,11 @@
         "import": "./dist/esm-browser/index.js",
         "require": "./dist/commonjs-browser/index.js"
       },
-      "default": "./dist/esm-browser/index.js"
+      "default": "./dist/esm-browser/index.js",
+      "types": {
```
SimenB, line 33: `types` should be first for some reason: https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing

package.json
```diff
@@ -29,7 +29,11 @@
...
+        "module": "./types.d.mts",
```
SimenB, line 34:
```suggestion
        "import": "./types.d.mts",
```

I've not seen `module` as an export condition before - could you point to its origin?

broofa: I believe this is [a webpack thing](https://webpack.js.org/guides/package-exports/#target-environment-independent-packages), possibly adopted by other packaging tools as well?

SimenB: Ah, cool. These are types tho, and the TS docs only point to node (where the spec comes from), and they don't have a `module` condition listed there: https://nodejs.org/api/packages.html#conditional-exports

types.d.ts
```diff
@@ -0,0 +1,241 @@
...
+ * }
+ *
+ * const v1Uuid = 'd9428888-122b-11e1-b85c-61cd3cbb3210';
+ * const v4Uuid = '109156be-c4fb-41ea-b1b4-efe1671c5836';
+ *
+ * uuidValidateV4(v4Uuid); // ⇨ true
+ * uuidValidateV4(v1Uuid); // ⇨ false
+ * ```
+ */
+export function validate(str: string): boolean
```
LinusU, line 224: In order to support #606:

```suggestion
export function validate(str: any): str is string
```

broofa:
Would it make sense to use JSDoc to document the actual source, then have `tsc` generate type definitions from that?

LinusU:
> * Does this solve the issue raised in [Have validate accept null and undefined #606](https://github.com/uuidjs/uuid/issues/606) which seems to exist in the current DefinitelyTyped types?

Left a note to myself to adress this 👍 

> * In the [current Types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/uuid/index.d.ts) we use rather complicated interfaces. What are the pros and cons of something like that?

I'm actually not sure why it was made that way 🤔 

It seems like it was introduced in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/16264.

@felipeochoa sorry for a ping for something you made 5 years ago 🙈, but do you happen to remember why use used:

```typescript
export type v1String = (options?: V1Options) => string;
export type v1Buffer = <T extends OutputBuffer>(options: V1Options | null | undefined, buffer: T, offset?: number) => T;
export type v1 = v1String & v1Buffer;
```

instead of using an overloaded function?

```typescript
export function v1(options?: V1Options | null): string
export function v1<T extends ArrayLike<number>>(options: V1Options | null | undefined, buffer: T, offset?: number): T
```

> * Also big +1 to auto-generating the docs from the types!

Awesome 🙌

LinusU:
> Would it make sense to use JSDoc to document the actual source, then have `tsc` generate type definitions from that?

This was actually my first approach, but I ran into a number of small issues. The biggest hurdle was probably that that would generate one `.d.ts` file per JavaScript-file, which then would have had to been copied to all the right places since we ship so many different versions 😅 It would be cool with something similar to Rollup but for `.d.ts` files!

If we go that route we could also run the TypeScript compiler to type check our JavaScript codebase with `--allowJs` and `--checkJs`. It looks at the JSDoc annotations and uses that to type check the code!

types.d.ts
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * @example
+ * ```javascript
+ * import { version as uuidVersion } from 'uuid';
+ *
+ * uuidVersion('45637ec4-c85f-11ea-87d0-0242ac130003'); // ⇨ 1
+ * uuidVersion('6ec0bd7f-11c0-43da-975e-2a8ad9ebae0b'); // ⇨ 4
+ * ```
+ */
+export function version(str: string): boolean
```
SimenB, line 241:
```suggestion
export function version(str: string): 1 | 3 | 4 | 5
```

or `number`? deffo not `boolean`, tho 🙂

types.d.ts
SimenB: nominal types would have made this way cleaner 😛

types.d.ts
LinusU: Hehe, not sure how it ended up as `boolean` 😅 

It can return `0` also, but yeah number union seems like a great return value, thanks! 👍

types.d.ts
LinusU: Yeah for sure!

types.d.ts
broofa: Hmm... worth adding a unit-test for types of some sort?  Jest + `tsc` a few sample scripts to verify it does / doesn't throw the expected type warnings?

SimenB: I'd recommend using `tsd` or something. In the jest repo we use https://github.com/jest-community/jest-runner-tsd which works great (if you already use jest), but tsd is great for type tests

LinusU:
I made some progress on keeping the docs in sync. The most things are already supported by [ts-readme-generator](https://github.com/LinusU/ts-readme-generator), and I've added some small missing things now.

There is still some things left though, the main thing being how to handle overloaded functions 😅 

But we only need a solution that's good enough for our specific needs, so that makes it a bit easier. I'm thinking now that it could compare if the overload only appends new arguments, and then merge them into just one heading. Should be quite easy!

I've put a todo list in the PR description to track what's left...

broofa:
> I've put a todo list in the PR description to track what's left...

Might want to add an item for removing `runmd` support.  ( remove from `devDependencies`, remove README_js.md)

LinusU:
@broofa ~I haven't looked closely at exactly how to integrate it, but `ts-readme-generator` only modifies the `## API` section of the readme, so I think that it can coexist with `runmd`. We could also add a flag to `ts-readme-generator` to just print the result to stdout, that way it could be integrated with `runmd` as well.~

Ahh, I see now that `runmd` is only used in the examples in the `## API` section 😅

types.d.ts
```diff
@@ -0,0 +1,241 @@
...
+ * [...bytes].map((v) => v.toString(16).padStart(2, '0')); // ⇨
+ *   // [
+ *   //   '6e', 'c0', 'bd', '7f',
+ *   //   '11', 'c0', '43', 'da',
+ *   //   '97', '5e', '2a', '8a',
+ *   //   'd9', 'eb', 'ae', '0b'
+ *   // ]
+ * ```
+ */
+export function parse(str: string): Uint8Array
```
ostera, line 38: What's the reasoning behind exposing the Uint8Array type like this? Could we instead return a value of type `Uuid` instead, to avoid leaking the fact that this is just bytes?

broofa: The `parse()` method is conceptually simple: Parse a UUID string into it's constituent bytes. This isn't going to change anytime soon.  `Uint8Array` is well understood to be the proper JS type for such things and is, in fact, what the method returns.

LinusU:
Note to self: change return value to `` `${string}-${string}-${string}-${string}-${string}` ``? 

ref: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/64798, https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1432

broofa:
> Note to self: change return value to `` `${string}-${string}-${string}-${string}-${string}` ``?

Is this a [Template Literal Type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)?  Seems like an appropriate change.

LinusU:
> > Note to self: change return value to `` `${string}-${string}-${string}-${string}-${string}` ``?
> 
> Is this a [Template Literal Type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)? Seems like an appropriate change.

Yes it is!

Yeah, I think it's a good change, since it would allow people to be stricter with functions that only accept uuids, and not any string 🚀 

-----

Also, sorry for falling behind on this, unfortunately I have a bit limited time at the moment, but I really want to push thru with this!

broofa:
FWIW, the more I think about this issue (adding type declarations to this project), the more I'm of the opinion that we should just port everything to TS, and let `tsc` generate the declarations for us.  Especially in light of the recent RFC9562 work.

Having this code be in TS would've given me more confidence that we're not introducing regressions.  It would also force us to be more strict about what types are supported (e.g. "What type, exactly, are binary UUIDs?").  And, of course, it'd avoid having to manually update the *.d.ts file here.

@pmccarren What's your comfort level with Typescript?  Would porting this codebase to TS help or hinder your interest in helping to maintain this project?

pmccarren:
@broofa I'm all in for typescript, and I believe it a worthwhile investment. I too feel it would build confidence working throughout the repo

As for comfort level, I've worked extensively with typed languages and while bit newer to the TS scene specifically, I've began adopting it exclusively for all new JS projects.

v10 focused on RFC9562, v11 TS + [deploy tooling](https://github.com/uuidjs/uuid/pull/681#issuecomment-2145723942)?

broofa:
> v10 focused on RFC9562, v11 TS + https://github.com/uuidjs/uuid/pull/681#issuecomment-2145723942?

Yup.  TS and deploy tooling our out of scope of v10.0.0 (which I'd like to release ASAP).

FWIW, I expect deploy tooling to be indepenent of semver releases.  (although I expect we'll want to do actual testing of the deploy pipeline, that can/should happen with prerelease tags that stay off the main release timeline.)  As for TS, switching to TS in and of itself wouldn't necessitate a major release since the API wouldn't necessarily change.  That said, it wouldn't surprise me if one of the side effects was tighter type restrictions in the API (e.g. `Uint8Array` adoption) that did break the API.

... but whatevs.  Cross those bridges post 10.0.0.

broofa:
Closing.  This PR seems to have eddied out (for now), and I think the path forward will be to port the `uuid` codebase to TS.

@LinusU Thanks for the effort.  Sorry we didn't get this merged.  So... uh... 'feel like porting some JS to TS??  😉

PR #655, opened 2022-10-12
export uuid regex

BrianHung:
Exports uuid regex. Useful when constructing other regexes that match or test strings containing uuids.

```ts
import { regex } from "uuid"
const cellPath = "cells/a87e96e8-51af-5383-bbf3-1797ebfeb41c"
const cellRegexp = new RegExp(`^cells/${regex.source}$`, "i");

for (let {op, path, value} of changes) {
  if (cellRegexp.test(path)) // ...
}
```


broofa:
This has been [discussed previously](https://github.com/uuidjs/uuid/issues/498).  tl;dr: everyone is going to want a slightly different flavor of regex (as your "[tweak](https://github.com/uuidjs/uuid/pull/655/files#diff-d77c8cadd4f5399649a8f5e0af42cd602058983d65e0ae07a94814d30329b7d1)" shows), and exposing the regex makes it harder for us to refine the behavior of `validate()` should we need to do that.

broofa:
(belated) ... but thanks for the submission.  Please don't let me discourage you from submitting PRs here or to other projects!  'Definitely appreciate the help.

BrianHung:
No worries!

PR #667, opened 2022-12-19
fix: small error

0xflotus:
I only fixed a small typo error


broofa:
Hey, thanks for the contribution.  [However ...](https://github.com/uuidjs/uuid/blob/main/README.md?plain=1#L1-L3)

0xflotus:
Ah thanks, I did the fix in the right file.

ctavan:
The generated readme doesn’t match expectations.

Can you ensure there the readme was generated using `npm run docs`?

0xflotus:
It's easier to open a new PR

Issue #580, opened 2021-08-09
Support new IETF UUID formats


dncnkrs:
A recently published draft RFC proposes three new time-based UUID formats. The UUIDv1 is also time-based, but the timestamp is formatted little-endian, whereas the new formats use a big-endian format. This means the UUID's will naturally sort in order of time, which has several engineering advantages.

Because the RFC is still a draft, I wouldn't yet expect that UUID libraries can generate the new formats. I'd like to know if this library intends to implement them, however.

I thought it would be useful to create this issue, so we can see if:

- the new formats will be implemented in this library,
- if so, the status of the implementation (assumed to be after the RFC is published). 

https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format-01

broofa:
Good issue.  I'm generally in favor of this, as the new versions have some much-needed improvements over the original spec.  However, this spec appears to still be an early, unadopted draft (e.g. numerous typos, not yet assigned an RFC #).

Until this is more formally adopted by the IETF, I don't think it's appropriate for us to officially support it.  (That's not to say we couldn't publish an "experimental" branch / release, but I don't see that as a high priority task.)

fabiolimace:
The new UUID formats are being discussed in this github repo: https://github.com/uuid6/uuid6-ietf-draft

kripod:
I've just created an isomorphic implementation for the current UUIDv7 draft: https://github.com/kripod/uuidv7 – I would appreciate some feedback about it.

CMCDragonkai:
@kripod in your example how does one provide the node id in a distributed context?

Also I was wondering if we can provide a custom clock source. In case we wanted a monotonic clock.

kripod:
@CMCDragonkai providing a custom machine ID isn’t possible yet, but 62 random bits should be sufficient to avoid collisions. Does the spec draft mention how to use machine IDs within a UUIDv7?

Not sure if a custom clock source may ever be provided, as JS is single-threaded and thus, `seq` is guaranteed to be monotonic.

Please open new issues about these at https://github.com/kripod/uuidv7/issues

CMCDragonkai:
Done:

* https://github.com/kripod/uuidv7/issues/1
* https://github.com/kripod/uuidv7/issues/2

CMCDragonkai:
[...]

Issue #636, opened 2022-05-19
Use `release-please` workflow for managing releases


broofa:
@ctavan: I've recently set up `runmd` to use https://github.com/google-github-actions/release-please-action for release handling.  'Haven't done a whole lot with it, but I've got it working and have tested it out.  It seems like a pretty decent solution to the problem of managing releases.

tl;dr: Any commit to `main` that indicates patch/fix/breaking change (via conventional commit message) will trigger the creation of a PR that updates CHANGELOG.md and package.json#version.   As new commits come in, the PR is updated to reflect the latest state of the release.  To push a release, just merge the PR (which triggers `npm publish` in the workflow.)

Seems like a pretty elegant solution.  I'm happy to put up a PR, but 'figured I'd float the idea here first, in case I'm overlooking something obvious.

ctavan:
Awesome! I like the idea that we can still control when to release through the CL. So feel free to introduce this here as well!

The biggest challenge I see at the moment though is that we would like to pack quite a few breaking changes into the next 9.0 release, but at least I personally hardly find spare time at the moment to invest into these changes… unfortunately as long as we don’t maintain several major version branches (which is also a hassle an likely doesn’t work well with automation) the fact that we have started merging these breaking changes into main also blocks merging non-breaking perf and cleanup fixes…. But again, this is something that hardly any tooling will solve for us.

broofa:
I think we can specify which branch(es) it runs on.  So we could have release PRs going for both the `main` and a `v9` branch.

ctavan:
After re-reading my previous comment I no longer agree with what I've written last time ;).

I think we should focus on the main branch and simply use semantic versioning to progress this project. The only event that I can think of where we would have to patch older major releases would be some sort of critical security vulnerability. And in that case I guess that irrespective of what automation we have in place, we'll have to find a manual way of doing so (which should also be no big deal with `git` & `npm publish`).

So if you could set this up for `uuid`, I think it would be an amazing improvement for the maintainability of this project!

PR #681, opened 2023-01-22
uuid v7

pmccarren:
### UUID V7

This PR implements UUID V7, ~closing~ ref #580

V7 is implemented, tested and documented and ready for review! No changes to the other versions were made.

Relevant RFC documents:

- [RFC 4122](https://www.ietf.org/rfc/rfc4122.html)
- [DRAFT of updates to RFC 4122](https://github.com/ietf-wg-uuidrev/rfc4122bis)
- [V7 Specification](https://ietf-wg-uuidrev.github.io/rfc4122bis/draft-00/draft-ietf-uuidrev-rfc4122bis.html#name-uuid-version-7)
- [Sample V7 value](https://ietf-wg-uuidrev.github.io/rfc4122bis/draft-00/draft-ietf-uuidrev-rfc4122bis.html#name-example-of-a-uuidv7-value)

### Benchmarks

V7 appears as fast as V4 when not using native crypto.randomUUID generation.

Results:

```
uuid.stringify() x 1,450,855 ops/sec ±2.12% (88 runs sampled)
uuid.parse() x 2,350,677 ops/sec ±1.19% (88 runs sampled)
---

uuid.v1() x 3,585,680 ops/sec ±0.86% (92 runs sampled)
uuid.v1() fill existing array x 9,229,031 ops/sec ±0.50% (93 runs sampled)
uuid.v4() x 13,834,376 ops/sec ±3.34% (84 runs sampled)
uuid.v4() fill existing array x 3,833,619 ops/sec ±1.51% (84 runs sampled)
uuid.v4() without native generation x 2,395,572 ops/sec ±1.27% (91 runs sampled)
uuid.v3() x 268,797 ops/sec ±1.32% (86 runs sampled)
uuid.v5() x 287,899 ops/sec ±1.12% (92 runs sampled)
uuid.v7() x 2,764,344 ops/sec ±0.95% (93 runs sampled)
uuid.v7() fill existing array x 2,801,443 ops/sec ±5.48% (84 runs sampled)
uuid.v7() with defined time x 3,223,107 ops/sec ±1.14% (89 runs sampled)
Fastest is uuid.v4()
```

TODO:
- [x] preserve monotonicity
- [ ] Add V7 type definitions to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/uuid)


broofa (commented):
Thanks for the contribution @pmmccarren.  This looks promising!

In addition to the inline comments, can you add a unit test alongside https://github.com/uuidjs/uuid/blob/main/test/unit/v1.test.js, with similar assertions.  Specifically, I'd like to see similar tests for sort order, uniqueness, and options handling.  (And anything else you think may be relevant.)

The sorting will be of particular interest, since producing UUIDS that sort the same whether sorted by timestamp or lexicographically was pretty much the whole point of version 7. 😄 

Note: This should not close #580, as the new RFC adds v6 and v8 formats, too, as well as the MAX_UUID constant.  v7 is definitely the majority of what's needed, but we'll want to keep 580 open until we fill in the remaining items.

Lastly, the new RFC is still making it's way through the review process.  I don't expect substantive changes at this point, so this is an appropriate time to be adding this. But as I noted in 580, I think it's prudent to go with an experimental version release tag of some sort.

@ctavan: Thoughts on how best to accept this?  Should we create a `v9-experimental` branch that we merge this into rather than `master`?  Then publish experimental releases from that?

README.md
```diff
@@ -248,6 +249,28 @@ import { v5 as uuidv5 } from 'uuid';
...
+| _returns_ | UUID `String` if no `buffer` is specified, otherwise returns `buffer` |
+
+Example:
+
+```javascript
+import { v7 as uuidv7 } from 'uuid';
+
+uuidv7(); // ⇨ '017f22e2-79b0-7cc3-98c4-dc0c0c07398f'
+```
+
```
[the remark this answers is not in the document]
pmccarren: Ah, appreciate the clarification! Should the generated `README.md` be included in the PR? I checked the GitHub workflows and only saw `docs:diff` being ran, but the changes seem to be discarded.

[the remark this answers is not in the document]
broofa: Yeah, add README.md changes to the PR.

src/v7.js
```diff
@@ -0,0 +1,54 @@
...
+  rnds[2] = (msecs / 0x1000000) & 0xff;
+  rnds[3] = (msecs / 0x10000) & 0xff;
+  rnds[4] = (msecs / 0x100) & 0xff;
+  rnds[5] = msecs & 0xff;
+
+  // set version (7)
+  rnds[6] = (rnds[6] & 0x0f) | 0x70;
+
+  // Per RFC4122 4.1.1, set the variant
+  rnds[8] = (rnds[8] & 0x3f) | 0x80;
```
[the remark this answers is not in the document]
pmccarren: nice catch, you're spot on. Reworked in similar nature to v1.

src/regex.js
```diff
@@ -1 +1 @@
-export default /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
+export default /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5,7][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
```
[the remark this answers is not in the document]
pmccarren: regex has been fixed to only pass validation for supported versions, and unit test coverage expanded to ensure only supported version numbers are validated (testing of failure cases).

pmccarren:
@broofa - glad to lend a hand! I appreciate your time reviewing the implementation.

This PR has been updated with your suggested changes, most notably:
- reworked buffer management so `rnds` is not mutated
- added v7 unit test coverage
- fixed README generation (via `npm run docs`)

In regards to sorting, I've been considering a few different approaches and while not presently implemented in this PR, I absolutely agree it should be :) I'll work on implementing this as well the associated unit tests.

src/regex.js
```diff
@@ -1 +1 @@
-export default /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
+export default /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[13457][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
```
LinusU, line 1: This accepted `2` before, nice catch 👍

src/v7.js
```diff
@@ -0,0 +1,55 @@
...
+  // [byte 7] - set 8 bits of random
+  b[i++] = rnds[7];
+
+  // [byte 8] - Per RFC4122 4.1.1, set the variant (2 bits)
+  b[i++] = (rnds[8] & 0x3f) | 0x80;
+
+  // [bytes 9-16] populate with random bytes
+  for (let n = 0; n < 7; ++n) {
+    b[i + n] = rnds[n + 9];
+  }
```
LinusU, line 50: Not sure if we want to change anything here, but unrolling the loop is _potentially_ faster and _potentially_ smaller when compressed with gzip.

```suggestion
  b[i++] = rnds[9];
  b[i++] = rnds[10];
  b[i++] = rnds[11];
  b[i++] = rnds[12];
  b[i++] = rnds[13];
  b[i++] = rnds[14];
  b[i++] = rnds[15];
```

note that V8 doesn't do loop unrolling: https://stackoverflow.com/a/66082249/148072

My guess for why it might be smaller when gzipped is because `];  b[i++] = rnds` will be repeated a lot, which should compress very well.

I don't think that it's a prerequisite for landing this, but I think it would be nice to benchmark and compare compressed size! If/when I get some time over I shall do this, since I think it's very interesting!

LinusU (commented):
Code looks good 👍 

> Lastly, the new RFC is still making it's way through the review process.

> Thoughts on how best to accept this? Should we create a v9-experimental branch that we merge this into rather than master? Then publish experimental releases from that?

Personally think that we should keep this PR open until the spec has been ratified. If we merge into a `v9` branch that would stop us from making a v9 release until that has happened, and that might potentially be a long time?

I think that it would be nice to backport the "version 2 is valid" bugfix right away to the current release though?

src/v7.js
LinusU: (a quick test of pasting just this function into [this online gzip compressor](https://facia.dev/tools/compress-decompress/gzip-compress/) gives 652B for the loop, and 640B when it's unrolled. This isn't running on the minified code though so is only indicative)

LinusU: A quick test seems to suggest that it's about 2.5% faster to unroll the loop in Chrome 109.0.5414.87 (Official Build) (arm64), and about the same (but the unrolled seems to always win, even if with a 0% margin) in Safari Version 16.2 (18614.3.7.1.5) (arm64).

https://jsbench.me/9vldb9u08i/1

ctavan:
Thanks an lot for the contribution!

I initially thought this new feature could land in the v9 release, but given we’re broadening the validation regex it’s a breaking change and we’ll have to bump the major version.

I think it would be nice to somehow release it as a prerelease so that people have a chance to try it and provide feedback, but I think our current release and branching process is not yet ready for multiple active major releases. If anyone’s willing to take a look at this I’m happy to support (we were considering better automation for this for a while… https://github.com/uuidjs/uuid/issues/636).

I agree however, that we should not release this in a public major release until the draft has been accepted.

src/regex.js
ctavan: I’m not sure about removal of version 2 from the regex.

v2 is clearly defined as a valid version in https://www.rfc-editor.org/rfc/rfc4122#section-4.1.3

While the RFC doesn’t provide any details how to generate such UUIDs and we don’t have support for them either, the version is still defined as valid and I think we should just stick to the RFC.

broofa (changes requested):

broofa:
Note: I've just created the `rfc4122bis` branch off `main` as a place to land this.  As @ctavan says, we haven't worked through the deploy & publish pipeline for non-`main` branches yet, so we'll have to figure that out.

Wish I could give you an ETA for that, but it's not been a priority for either of us.  As he says... "Help Wanted".

src/regex.js
LinusU: Missed that, yeah makes sense to keep it 👍

src/v7.js
LinusU: > However in this case I kind of like how the unrolled code reads. It flows well with the other `b[i++]` assignments in prior bytes, so I'm good with either form.

Yeah I agree, that was actually what got me started on this, hehe 😁

pmccarren: @LinusU I agree, the unrolled code reads more clearly and fits cleanly with the other byte manipulations. As an aside, I enjoyed going through your performance breakdown - learned a few things!

src/regex.js
pmccarren: @ctavan @broofa as this PR is headed towards encompassing the broader RFC updates (v6 and v8 included) and likely to be merged when the RFC is ratified, I see it worthwhile to validate them as well. Any objections to the following?

```suggestion
export default /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-8][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
```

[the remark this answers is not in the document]
pmccarren: marking conversation resolved; duplicate of https://github.com/uuidjs/uuid/pull/681#discussion_r1086227990

broofa: Yup.  Go ahead and commit.

broofa (approved):
@ctavan @LinusU Any other comments?  If not, I'd like to get this merged (pending @pmccarren commiting his suggested change for the regex to allow `1-8`).

Regarding the plan for the `rfc4122bis` branch, I've created a [project to capture the remaining work](https://github.com/orgs/uuidjs/projects/1/views/1), with brief notes on possible implementation details.

Honestly, there's not all that much left... if anyone is feeling inspired.  ;-)

Lastly, thoughts on inviting Patrick to join the UUIDJS org?  He's done some solid work here that's much appreciated, and seappears to have good street cred.  If people could weigh in on that idea (including you, Patrick), we can move forward or not as appropriate.  [cc'ing @TrySound here]

pmccarren:
@broofa just merged the `1-8` regex update. I'm close to done with implementing monotonic generation - I'd prefer to merge once that's wrapped. Will be buttoned up in a day or two.

As for UUIDJS Org, I'll let my work speak for itself but will add that if you'll have me I'll be an active participant :)

pmccarren:
@broofa I just finished adding monotonic support, and it's ready for review! Took a bit of finesse but I'm rather happy with how it turned out. In addition to the monotonicity related unit tests, I manually tested sorting preservation during generation of 100M uuids.

couple of sequence counter implementation notes:
* when additional generations occur inside the same millisecond, we use a dedicated 31 bits as an incrementing sequence counter. Referred to as _Fixed-Length Dedicated Counter Bits (Method 1)_ [in the draft RFC](https://ietf-wg-uuidrev.github.io/rfc4122bis/draft-00/draft-ietf-uuidrev-rfc4122bis.html#section-6.2-5.2.1).
* seq is 31 bits and (re)initialized from the random data pool whenever the clock advances/changes. as the seq is initialized randomly, for nominal usage there is a significant amount (74 bits) of randomness in a given v7 uuid.
* when the sequence counter rolls over we increment the internal date by 1 millisecond and continue. this accepts the tradeoff of minor clock drift for lexicographical sorting in substantial batch id generation workloads.
* if the internal date ever exceeds 10 seconds beyond system time, both the date and seq are hard reset.
* lexicographical sorting is preserved up to `(2^31)*10000` generations for a provided millisecond.
* the seq is stored as 31 bits. the 12 high bits and 19 low bits are stored separately in the uuid to preserve sorting while maintaining a large enough counter size

Pending review, I believe this PR is ready to go now!

pmccarren:
CI failures tracked in #688 (actions/setup-node npm 9)

robinpokorny:
Hey, @pmccarren, I had a quick look and have some questions. If you have time, could you look at them?

1. Should there be more information that the code includes monotonic guarantee beyond the timestamp? While I think it's a good default, it seems fair to highlight it.
1. Should there be an option to opt-out of that? That is, make the `seq` random for each call even in the same millisecond. (In this case, I was thinking about `updating` a v4 UUID with timestamp and version, while potentially reusing native v4 generation.) 
1. Does the fact that we always increase the `seq` by one, even for the first call within a millisecond, have an impact on the randomness?
1. Why did we choose 10 second? That seems like a lot.
1. What about mixed use of user providing `msecs` or not (as this library is used a lot, it can happen quite easily)? It seems it would break the monotonic guarantee. Is that a concern?

src/regex.js
rllane17: T

wiperawa:
Just wondering when approx. this PR will be merged?

broofa:
@wiperawa No specific ETA.  I just need to find the time to review and work through deploy process for an experimental branch.

FWIW, we (CodePen) actually have a need for this as well, so I do have a bit of incentive to make this happen.   I just need to find the time to sit down and do a proper review.

(And continued apologies to @pmccarren for dragging our feet on this.  This is still important work that we intend to merge.)

rdrpenguin04:
One of my teams is also waiting on UUID v7 for their project; we'd also greatly appreciate if this were reviewed, especially since it already has one approval.

claytongulick:
Eagerly awaiting this to be released as well.

Thanks for the great library and work on it!

README.md
```diff
@@ -17,7 +17,9 @@ For the creation of [RFC4122](https://www.ietf.org/rfc/rfc4122.txt) UUIDs
 - **Small** - Zero-dependency, small footprint, plays nice with "tree shaking" packagers
 - **CLI** - Includes the [`uuid` command line](#command-line) utility
 
-**Upgrading from `uuid@3`?** Your code is probably okay, but check out [Upgrading From `uuid@3`](#upgrading-from-uuid3) for details.
+> **Note** Upgrading from `uuid@3`? Your code is probably okay, but check out [Upgrading From `uuid@3`](#upgrading-from-uuid3) for details.
+
+> **Note** Only interested in creating a version 4 UUID? You might be able to use [`cypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID), eliminating the need to install this library.
```
ihmpavel, line 22: Looks like a small typo here

```suggestion
> **Note** Only interested in creating a version 4 UUID? You might be able to use [`crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID), eliminating the need to install this library.
```

README_js.md
```diff
@@ -32,7 +32,9 @@ For the creation of [RFC4122](https://www.ietf.org/rfc/rfc4122.txt) UUIDs
 - **Small** - Zero-dependency, small footprint, plays nice with "tree shaking" packagers
 - **CLI** - Includes the [`uuid` command line](#command-line) utility
 
-**Upgrading from `uuid@3`?** Your code is probably okay, but check out [Upgrading From `uuid@3`](#upgrading-from-uuid3) for details.
+> **Note** Upgrading from `uuid@3`? Your code is probably okay, but check out [Upgrading From `uuid@3`](#upgrading-from-uuid3) for details.
+
+> **Note** Only interested in creating a version 4 UUID? You might be able to use [`cypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID), eliminating the need to install this library.
```
ihmpavel, line 37: Another small typo here

```suggestion
> **Note** Only interested in creating a version 4 UUID? You might be able to use [`crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID), eliminating the need to install this library.
```

mbrimmer83:
Rumor has it uuid v7 is now part of the [proposed standard](https://www.rfc-editor.org/info/rfc9562) What would it take to get this over the finish line?

ryan-keswick:
Any updates on this?

broofa:
Did another review pass.  Pushed a couple minor documentation tweaks to update things now that RFC9562 is official.

@ctavan:  In the interests of moving forward with 9562, I'm merging this.  If there are questions or concerns, we can address those in followup PRs.

@pmccarren:   Thanks for the great work on this.  My apologies for letting this stagnate for so long.  Now that 9562 is official, my goal is to get this pushed out ASAP.

ctavan:
I‘m all for it, thanks for taking care!

But let’s make use of the conventional commit message format in the squash commits so we get proper autogenerated changelogs.

broofa:
> let’s make use of the conventional commit message format in the squash commits

Ugh... I keep screwing that up. :-(

ctavan:
Does GitHub have an option to run actions (commitlint) before squash merges?

broofa:
There's probably a way, but... 🤷

pmccarren:
@broofa and @ctavan I appreciate your time in reviewing and assisting with the PR. I'm excited we're collectively one step closer to uuid7 being widely available :)

If there is a need for assistance, I have time I can dedicate towards experimental deploys. Just let me know if it'll be helpful, and I'm happy to dive in.

broofa:
@pmccarren That'd be awesome.  On the deploy front, there's two nuts we'd like to crack:
1. Enforce Conventional Commits™ on every commit that goes onto the `main` branch.  (witness my fumbling the message on this PR)
2. Automate the deploy process so we're not manually publishing from our laptops.

We haven't tackled the former yet, but if you know a way to do that, we're all for it.

For the latter, I put up #725 a couple years ago before getting distracted by other things.  The main goal there was to get away from our current ad-hoc release process that has us creating PRs just to update `package.json#version`, and running `standard-version` and `npm publish` manually on our laptops.  The goal there was to automate the folllowing steps:
  * Update CHANGELOG with notes for all relevant changes
  * Bump `package.json#version` based on Conventional Commit message semantics
  * Create git tag for release
  * Run `npm publish` in a workflow

`release-please` is (I think) still a good choice for that but I haven't followed development.  If you can get that working, that'd be great.  Or if there's another option you think would work better I'm open to suggestions.

I've sent you an invite to the `uuidjs` org here.   Let me know if there are permissions you need here or on NPM for this work.  (Obviously we'll need to be careful about what gets npm-published.)

And thank you for stepping up to help (again).

sachinjnd:
@broofa / @pmccarren Now that the changes for uuid v7 are merged, what's the plan for release on npm?

ctavan:
See #760

Issue #688, opened 2023-02-18
Fix CI workflows and add support for npm 9


pmccarren:
Hello! Our CI pipeline is a bit grumpy at present, and this issue works to provide a remedy.

Github actions previously ran fine, however recently started failing with no change to the repo. Example of recent [successful run](https://github.com/uuidjs/uuid/actions/runs/4089640855/jobs/7056335853#step:7:9), and [failing run](https://github.com/uuidjs/uuid/actions/runs/4100489497/jobs/7259834200#step:7:9).

This failure is due to `actions/setup-node` now installing `npm 9.x` by default with `node 18.x`, _instead of_ the previous `npm 8.x`. We could apply a bandaid by pinning npm 8 in CI, but it prudent to support npm 9.

npm 9 introduces several significant [breaking changes](https://docs.npmjs.com/cli/v9/using-npm/changelog#-breaking-changes).

This issue intends to track the various tasks needed to resolve these issues.
- [x] `npm bin` was removed, add a workaround in `scripts/build.sh`
- [ ] fix npm install inside examples directory. This one is tricky. I have not found any noted change to how npm 9 handles symlinks, but our `.local/uuid` pattern no longer works as expected, resulting in an npm error during npm install.

ctavan:
I think this was fixed by #695. Please reopen if not.

PR #689, opened 2023-02-18
fix: CI node 18.x

pmccarren:
This PR contains changes resolving `actions/setup-node` `node 18.x` runs, and adds support for `npm 9`.

Closes: #688

TODO:
- [x] `npm bin` was removed, add a workaround in `scripts/build.sh`
- [ ] fix npm install inside examples directory. This one is tricky. I have not found any noted change to how npm 9 handles symlinks, but our `.local/uuid` pattern no longer works as expected, resulting in an npm error during npm install.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -4,7 +4,7 @@
 ROOT="$(pwd)/$(dirname "$0")/.."
 cd "$ROOT" || exit 1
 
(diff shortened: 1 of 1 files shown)


ctavan:
I think this was fixed by #695.

PR #699, opened 2023-03-21
chore: upgrade and fix the stale-issues workflow

ctavan:
This has been failing with the following error message for a while:

```
The workflow is not valid. .github/workflows/stale-issues.yml (Line: 20,
Col: 32): A sequence was not expected .github/workflows/stale-issues.yml
(Line: 21, Col: 29): A sequence was not expected
```


LinusU (approved):
LGTM 👍

PR #700, opened 2023-03-22
chore: Add Dependabot for dependency updates

nschonni:
Noticed some of the GitHub Actions were a few major versions behind. Instead of a one-off fix, thought Dependabot made more sense to keep them up-to-date


.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
```
broofa, line 5: Comment not needed.  Link is broken.
```suggestion
# To get started with Dependabot version updates, you'll need to specify which
# package ecosystems to update and where the package manifests are located.
# Please see the documentation for all configuration options:
# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates

```

nschonni:
> Would it make more sense to enable this in `Settings -> Code security and analysis`? (Are these doing different things? If so, in what way?)

The one in the repo settings is only for security related issues. EX: it will create pull requests if there is a vulnerable version of ESLint, but not when there is a new version of ESLint to keep it up-to-date

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: '/'
+    schedule:
+      interval: weekly
```
nschonni, line 6: Can be increased to Monthly, if you think weekly will be too noisy
```suggestion
      interval: monthly
```

.github/dependabot.yml
nschonni: Removed the block. I copied it from another repo, which must have been from an older generated file

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: npm
```
ctavan, line 3: I'm not convinced. All npm dependencies in this repo are exclusively for the build and test toolchain. There's only really a need to upgrade whenever we're looking into releasing a new version of the library, which happens very rarely. I think this will create quite a bit of noise at no clear benefit.

I think even running this monthly would create an amount of noise that I'm personally currently not willing to handle.

nschonni: Do you want me to remove that one, but leave the GitHub Actions part?

broofa:
@nschonni I'm on the fence about this as well.  Let's table this for now.  As @ctavan says, I don't know that this is a problem we really need to be solving at this time.  Updating manually as-needed isn't that big a deal.

I do appreciate your taking the time to bring this to our attention, though.

PR #703, opened 2023-03-23
ci: run browser tests on pull_request_target

nschonni:
`pull_request_target` allows runs to access the secret values. Since this adds an attack vector described here https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ I added the extra condition to only run on the main repo `push` or if someone adds a `safe to test` label after reviewing the code

### DIFF
--- a/.github/workflows/browser.yml
+++ b/.github/workflows/browser.yml
@@ -1,25 +1,28 @@
 name: Browser
 
-on: [push]
+on:
+  push:
+  pull_request_target:
+    types: [labeled]
 
 jobs:
   browser:
-
+    if: github.repository == 'uuidjs/uuid' && (contains(github.event.pull_request.labels.*.name, 'safe to test') || github.event_name == 'push')
     runs-on: ubuntu-latest
     timeout-minutes: 30
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/browser.yml
```diff
@@ -1,25 +1,28 @@
...
-      with:
-        fetch-depth: 10
+    - uses: actions/checkout@v3
     - name: Use Node.js 16.x
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v3
       with:
         node-version: 16.x
-    - run: npm install
+    - run: npm ci
```
ctavan, line 20: Great improvement, I didn't know [`npm ci`](https://docs.npmjs.com/cli/v9/commands/npm-ci) before.

ctavan (approved):
Thank you for this contribution. I think the security risk in this case is acceptable.

broofa:
Question: What happens if a PR is tagged "safe to test" and then subsequently modified by the author (e.g. in some malicious way)?

(Not questioning this PR.  Just a general question I've had for awhile.)

PR #705, opened 2023-03-23
chore: adapt bundlewatch config to new main branch name

ctavan:


broofa (approved):
Was this not working previously?  (Sorry, haven't been paying that close attention to this side of things, obviously. 😳 )

ctavan:
I think it was still comparing the bundlesize against the last known values from the former `master` branch but not recording new values when running on `main` (after a PR merge) because `main` was not in the `trackedBranches`…

PR #709, opened 2023-04-12
test: remove missing getRandomValues test

### LINKED ISSUE #708: [BUG] The test failed on the latest node version `rng` `Browser without crypto.getRandomValues()'` test fail
### Before you begin...

- [X] I have searched the existing issues
- [X] I am not using version 13.x of node  (if so, please upgrade)

### Description of the problem

The test is failing on the latest node version (19.9.0):
https://github.com/uuidjs/uuid/blob/6eef540aa3407b15e1e7573f45ff17098e9343ea/test/unit/rng.test.js#L15-L19

as crypto is [globally available](https://nodejs.org/api/globals.html#crypto:~:text=v19.0.0-,No%20longer%20behind%20%2D%2Dexperimental%2Dglobal%2Dwebcrypto%20CLI%20flag.,-v17.6.0%2C%20v16.15.0), which causes the CITGM tests in node to fail.

related PR:
- https://github.com/nodejs/node/pull/47413

### Recipe for reproducing

run:

```bash
nvm install 19.9.0
nvm use 19.9.0
npm test
```

### Additional information

_No response_

### Environment

_No response_

LinusU:
Fixes #708

### DIFF
--- a/test/unit/rng.test.js
+++ b/test/unit/rng.test.js
@@ -1,6 +1,5 @@
 import assert from 'assert';
 import rng from '../../src/rng.js';
-import rngBrowser from '../../src/rng-browser.js';
 
 describe('rng', () => {
   test('Node.js RNG', () => {
@@ -10,12 +9,6 @@ describe('rng', () => {
     for (let i = 0; i < bytes.length; ++i) {
       assert.equal(typeof bytes[i], 'number');
     }
-  });
-
-  test('Browser without crypto.getRandomValues()', () => {
-    assert.throws(() => {
-      rngBrowser();
-    });
   });
 
   // Test of whatwgRNG missing for now since with esmodules we can no longer manipulate the


test/unit/rng.test.js
```diff
@@ -12,7 +12,8 @@ describe('rng', () => {
     }
   });
 
-  test('Browser without crypto.getRandomValues()', () => {
+  // Newer versions of Node.js have a global crypto object
+  (global.crypto == null ? test : xtest)('Browser without crypto.getRandomValues()', function () {
```
ctavan, line 16: Thank you for looking into this!

I’m not quite sure though if this change is so useful. I think originally we tried to model the case where no global crypto object is present in which case we expect the library to throw in browser environments…

Is there any way to fake such an environment in the node test suite, e.g. by somehow unsetting the global crypto object?

Otherwise I‘m wondering if we should remove this test case altogether since we don‘t provide any polyfills for old browsers anyways and browsers without crypto (that was only IE<10) don’t play a role in practice anymore.

ctavan (changes requested):

broofa (approved):

ctavan (approved):

PR #718, opened 2023-05-23
Support "react-native" package entry point

liamjones:
Adds conditional export for `react-native` environment.

This fixes UUID not working in Jest with the `react-native` preset. Before this change it was falling back to default and then failing to find the global `crypto`.

`react-native` inserted after `browser` so React Native for Web can still be targeted properly, see https://nodejs.org/api/packages.html#community-conditions-definitions for more info.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,11 @@
         "import": "./dist/esm-browser/index.js",
         "require": "./dist/commonjs-browser/index.js"
       },
+      "react-native": {
+        "module": "./dist/esm-node/index.js",
+        "require": "./dist/index.js",
+        "import": "./wrapper.mjs"
+      },
       "default": "./dist/esm-browser/index.js"
     },
     "./package.json": "./package.json"


liamjones:
Oh, and I should have explained, since RN 0.71 there's a custom environment for Jest which extends `NodeEnv`: https://github.com/facebook/react-native/commit/cb2dcd327c62edae58bb8571273add4c286fd2a8#diff-70258ed063b4b3f0c6a81e698a3f928cd310f4782e8c260e08900809f5cfa9e4

broofa:
> there's a custom environment for Jest which extends NodeEnv

Naive question:  Is this an issue with that custom environment, then?  E.g. the "react-native" configuration you've added is identical to the ["node" configuration](https://github.com/uuidjs/uuid/pull/718/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R27-R31).  Could `jest` not fallback to the "node" configuration here in that case?

(Asking because I fundamentally believe that adding complexity to address context-specific issues in downstream dependencies is a pretty serious code smell.  We do it, but mostly because @ctavan is a much more understanding and gracious maintainer than I am. :-) )

Also, would it be possible to add a test-case for this?  I'm not familiar with react-native so I don't know what that would look like, but it would be helpful to have code that demonstrates the problem.

liamjones:
>  Could jest not fallback to the "node" configuration here in that case?

Interesting question! I don't know, let me do some more research on the RN/Jest side and see if what you've suggested is possible...

liamjones:
For anyone finding this issue in future, the relevant PR on the RN project is here: https://github.com/facebook/react-native/pull/39237

PR #722, opened 2023-07-25
Fix problems with jest

alfonsocv12:
Only fix problems with jest

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,9 +1,14 @@
-export { default as v1 } from './v1.js';
-export { default as v3 } from './v3.js';
-export { default as v4 } from './v4.js';
-export { default as v5 } from './v5.js';
-export { default as NIL } from './nil.js';
-export { default as version } from './version.js';
-export { default as validate } from './validate.js';
-export { default as stringify } from './stringify.js';
-export { default as parse } from './parse.js';
+import { default as v1 } from './v1.js';
+import { default as v3 } from './v3.js';
+import { default as v4 } from './v4.js';
+import { default as v5 } from './v5.js';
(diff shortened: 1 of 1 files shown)


walis85300 (approved):

moisesmorillo (approved):

broofa:
Please elaborate on what problem this is fixing.  Preferably with a minimal, verifiable test case.

alfonsocv12:
I'm using AWS library that depends on this library, my setup is Jest with Typescript, and this lines are giving syntax errors, with that setup specifically.

I know that this works perfectly with vanilla JS, but changing the syntax into this will give support to more use cases.

here is the error message, that I'm getting.

BTW. we fix the issue at our end by ignoring this file

Screenshot 2023-08-20 at 3 39 46 PM

ctavan:
Are you building for Node.js or for the browser?

alfonsocv12:
Node.js

ctavan:
> Node.js

Ok, this looks like some configuration issue in the build chain then: as you can see from the error message, it tries to load the `esm-browser` build of the library, however under Node.js it shouldn’t be loading the browser build.

I would recommend checking the typescript and jest setup for anything related to exports and module handling. You can check other jest related bugs in this repo for inspiration: https://github.com/uuidjs/uuid/issues?q=is%3Aissue+is%3Aclosed+jest

ctavan:
In addition, providing a [minimal
reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) would greatly help us PinPoint and understand the issue.

broofa:
No recipe for reproducing provided.  Closing.

PR #725, opened 2023-08-24
build: Enable release-please workflow for release, fixes #636 

### LINKED ISSUE #636: Use `release-please` workflow for managing releases

broofa:
See https://github.com/googleapis/release-please.

tl;dr: New commits will trigger the creation of a "release PR".  Merging the release PR will automatically create a new release.

I'm not super-familiar with this, so will probably be flailing around a bit in this PR as I figure out what's needed.  And, alas, knowing *for sure* that it's working requires actually publishing releases.


ctavan:
Awesome!

I think for this to be maximally useful we need to somehow enforce that the commit messages of our squash commits that go into the main branch follow the conventional commit format.

I think we do have a presubmit that checks this when committing locally, but this is not run upon squash merging a PR.

Any idea how we could enforce properly formatted messages for these squash merges?

broofa:
> I think for this to be maximally useful we need to somehow enforce that the commit messages of our squash commits that go into the main branch follow the conventional commit format.

I don't have a good solution to that at the moment.  ... but that's also a separate issue from this PR, right?

FWIW, I'm bogged down at the moment on the lack of support for prerelease versions in `release-please`.  I'd really like to get this setup for the `rfc4122bis` branch so we can publish to NPM using "9.0.0-bis.#"-style versions, but blocked by [this issue](https://github.com/googleapis/release-please/issues/510).  (I'm not well-versed in how release-please works, though, so there may be a workaround...?)

ctavan:
> > I think for this to be maximally useful we need to somehow enforce that the commit messages of our squash commits that go into the main branch follow the conventional commit format.
> 
> I don't have a good solution to that at the moment. ... but that's also a separate issue from this PR, right?

Correct, it‘s an independent issue. However I‘m wondering if we should fix the conventional-commit enforcement first? I think release automation can only reliably pick the right major/minor/patch version to increment if the commit messages obey the conventional commit rules. Otherwise we may end up with release-please creating minor bumps for breaking changes. OTOH it only generates a pull request so we still have human supervision. Given that, I think the order in which we do those two things actually doesn’t matter.

hexiaokang (approved):

broofa:
@Shooteger: No, this about improving the release process in general, not about pushing out a specific release.

broofa:
@ctavan @pmccarren FWIW, the `standard-version` (that we depend on) was deprecated last year and now recommends using `release-please` instead: https://github.com/conventional-changelog/standard-version

broofa:
'Not really sure how well this is going to work but this PR is long in the tooth and the way release-please works makes it difficult to properly test on a branch.  So I'm merging and will work the kinks out as part of rolling `v11`.

PR #732, opened 2023-09-07
ci: update node versions for cI

broofa:


ctavan (approved):

ctavan:
I updated the branch protection to expect the 20.x check to pass instead of 12.x

ctavan:
Uhm, maybe we should add back the 12.x test? Otherwise I think it would be technically a breaking change to release a new version, WDYT?

We should also update the README for Node 20.x support…

broofa:
> Uhm, maybe we should add back the 12.x test? Otherwise I think it would be technically a breaking change to release a new version, WDYT?

Hmm... good point.  I'll add 12.x back in.  

[Breadcrumb: [link to the debate on whether this sort of thing constitutes a breaking change](https://github.com/semver/semver/issues/716#issuecomment-985928530).]

That said, our node support policy is "supported node releases + 1 prior version", right?  So with 16.x slated for [premature](https://nodejs.org/en/blog/announcements/nodejs16-eol) end-of-life on Monday, we should be able to drop support for 12 & 14 at our leisure.

PR #750, opened 2024-06-03
feat!: update node support matrix (only support node 16-20)

broofa:
This will be a breaking change.  My intention is to release `uuid@10` release with RFC9562 support, so this is a good opportunity to update the CI matrix.


ctavan (approved):

ctavan:
Github settings for the main branch also still need to be updated.

PR #752, opened 2024-06-03
chore: update package dependencies

broofa:
Updating all out-of-date `[dev]dependencies`.  The two major ones are `eslint` and `prettier`. 

This also switches to `eslint`'s flat config file format, which is always a "fun" exercise.  On this occasion, because we're using StandardJS, I ran into [this issue](https://github.com/standard/eslint-config-standard/issues/411).  And I guess there's some opensource [clownery going on](https://github.com/standard/standard/issues/1948#issuecomment-2138078249) in that project where the current published version of `eslint-config-standard`  is broken but there's some debate over governance or... something?  I dunno, I spent way too much time trying to figure out what was broken, and then why it can't be fixed, and then why the project had to be forked, then trying to figure out how to get the fork of the project working here.

And if anyone from StandardJS or NeoStandard ends up reading this... I WANT THE LAST TWO HOURS OF MY LIFE BACK, PLEASE!


package.json
```diff
@@ -17,9 +17,7 @@
     "rfc4122"
   ],
   "license": "MIT",
-  "bin": {
-    "uuid": "./dist/bin/uuid"
-  },
+  "bin": "./dist/bin/uuid",
```
ctavan, line 20: What’s the effect of this change?

ctavan (approved):

package.json
broofa: Nothing, since [both forms are allowed](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin).

I'm not actually sure why this changed.  I noticed it eariler and assumed it was eslint or prettier being clever, but that doesn't appear to be the case. 🤷   I'll revert since the docs kind of make it sound like the object-form is preferred.

voxpelli:
_/me grants @broofa two extra hours of life and asks for forgiveness – will try to make up for it_

Issue #651, opened 2022-08-24
[BUG] importing UUID causes an 'Element type is invalid' error


BeetrootShoulders:
### Before you begin...

- [X] I have searched the existing issues
- [X] I am not using version 13.x of node  (if so, please upgrade)

### Description of the problem

Importing uuid into a Next.js component using the method in the docs:

```
import { v4 as uuidv4 } from "uuid";
```
causes an error:

```
Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

### Recipe for reproducing

```bash
Import UUID into a Next component.
```

### Environment

```
System:
    OS: Windows 10 10.0.22000
    CPU: (12) x64 Intel(R) Core(TM) i5-10400 CPU @ 2.90GHz
    Memory: 1.33 GB / 7.75 GB
  Binaries:
    Node: 16.14.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.5.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (104.0.1293.63)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    @babel/preset-react: ^7.18.6 => 7.18.6 
    @sanity/client: ^3.3.3 => 3.3.3 
    @sanity/image-url: ^1.0.1 => 1.0.1 
    autoprefixer: ^10.4.8 => 10.4.8 
    eslint: 8.22.0 => 8.22.0 
    eslint-config-next: 12.2.5 => 12.2.5 
    moment: ^2.29.4 => 2.29.4 
    next: 12.2.5 => 12.2.5 
    postcss: ^8.4.16 => 8.4.16 
    react: 18.2.0 => 18.2.0 
    react-dom: 18.2.0 => 18.2.0 
    tailwindcss: ^3.1.8 => 3.1.8 
    uuid: ^8.3.2 => 8.3.2 
```

BeetrootShoulders:
Very helpful. That's as minimal and complete as it gets. I do nothing but import uuid into a component, do nothing else with it, and the error gets thrown.

broofa:
https://codesandbox.io/s/next-js-forked-odu8zx?file=/index.js

[...]

PR #759, opened 2024-06-05
feat: v8 support

broofa:
[Stacking this on top of #754, which I expect will go in shortly]

We actually snuck in the change to allow version 8 in the validation regex as part of #651, which is really all that's required to support v8 UUIDs.  There's no formal algorithm for v8 ("experiment or vendor-specific") ids, so the only impact on this library is that such ids must properly `validate()` and `version()`.

This PR adds tests for those cases.  I've also taken the opportunity to refactor the `validate` and `version` tests to be "table-driven" from a common set of test cases, which I've fleshed out to include the original cases we had, plus a few additional ones.


ctavan (approved):

pmccarren:
@broofa I was thinking it might be helpful to provide an API for generating a uuid8, such as `v8(buf)`. The method would manage bit shifting / placement, and return a compliant uuid8 populated with the first 122 bits of the buffer. Thoughts on wether or not this utility belongs?

pmccarren (approved):

broofa:
> @broofa I was thinking it might be helpful to provide an API for generating a uuid8, such as v8(buf). The method would manage bit shifting / placement, and return a compliant uuid8 populated with the first 122 bits of the buffer. Thoughts on wether or not this utility belongs?

Let's hold off on this until we see demand for it.  I suspect v8 UUIDs will generally be a pretty niche use case compared to other versions, and experience has taught me to be cautious about expanding the API surface.  It's 10x harder to remove an API than it is to add it.

ctavan (approved):
Great refactor of the tests, thank you!

PR #738, opened 2023-10-11
fix: revert "perf: remove superfluous call to toLowerCase (#677)"

broofa:
This reverts commit e53793f5be93413cad68b05b3a339ca23e993342 per [this discussion](https://github.com/uuidjs/uuid/pull/677#issuecomment-1757351351)

PR #760, opened 2024-06-07
chore(release): 10.0.0

broofa:
Manually edited the CHANGELOG to include "uuid v7" PR.

@pmccarren @ctavan: This is the last step before `npm publish`.  Speak now or hold your peace. :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [10.0.0](https://github.com/uuidjs/uuid/compare/v9.0.0...v10.0.0) (2024-06-07)
+
+### ⚠ BREAKING CHANGES
+
+- update node support (drop node@12, node@14, add node@20) (#750)
+
+### Features
+
+- support support rfc9562 MAX uuid (new in RFC9562) ([#714](https://github.com/uuidjs/uuid/issues/714)) ([0385cd3](https://github.com/uuidjs/uuid/commit/0385cd3f18ae9920678b2849932fa7a9d9aee7d0))
+- support rfc9562 v6 uuids ([#754](https://github.com/uuidjs/uuid/issues/754)) ([c4ed13e](https://github.com/uuidjs/uuid/commit/c4ed13e7159d87c9e42a349bdd9dc955f1af46b6))
+- support rfc9562 v7 uuids ([#681](https://github.com/uuidjs/uuid/issues/681)) ([db76a12](https://github.com/uuidjs/uuid/commit/db76a1284760c441438f50a57924b322dae08891))
+- update node support matrix (only support node 16-20) ([#750](https://github.com/uuidjs/uuid/issues/750)) ([883b163](https://github.com/uuidjs/uuid/commit/883b163b9ab9d6655bfbd8a35e61a3c71674dfe1))
+- support rfc9562 v8 uuids ([#759](https://github.com/uuidjs/uuid/issues/759)) ([35a5342](https://github.com/uuidjs/uuid/commit/35a53428202657e402e6b4aa68f56c08194541bf))
+
+### Bug Fixes
+
+- revert "perf: remove superfluous call to toLowerCase ([#677](https://github.com/uuidjs/uuid/issues/677))" ([#738](https://github.com/uuidjs/uuid/issues/738)) ([e267b90](https://github.com/uuidjs/uuid/commit/e267b9073df1d0ce119ee53c0487fe76acb2be37))
+
 ## [9.0.1](https://github.com/uuidjs/uuid/compare/v9.0.0...v9.0.1) (2023-09-12)
 
 ### build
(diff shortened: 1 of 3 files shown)


BitWonka:
came here to check up on status of uuid v7 and now only have to look for LGTM!

CHANGELOG.md
```diff
@@ -2,6 +2,24 @@
...
 
+## [10.0.0](https://github.com/uuidjs/uuid/compare/v9.0.0...v10.0.0) (2024-06-07)
+
+### ⚠ BREAKING CHANGES
+
+- update node support (drop node@12, node@14, add node@20) (#750)
+
+### Features
+
+- support support rfc9562 MAX uuid (new in RFC9562) ([#714](https://github.com/uuidjs/uuid/issues/714)) ([0385cd3](https://github.com/uuidjs/uuid/commit/0385cd3f18ae9920678b2849932fa7a9d9aee7d0))
```
ctavan, line 13: Support support?

CHANGELOG.md
```diff
@@ -2,6 +2,24 @@
...
+- support rfc9562 v6 uuids ([#754](https://github.com/uuidjs/uuid/issues/754)) ([c4ed13e](https://github.com/uuidjs/uuid/commit/c4ed13e7159d87c9e42a349bdd9dc955f1af46b6))
+- support rfc9562 v7 uuids ([#681](https://github.com/uuidjs/uuid/issues/681)) ([db76a12](https://github.com/uuidjs/uuid/commit/db76a1284760c441438f50a57924b322dae08891))
+- update node support matrix (only support node 16-20) ([#750](https://github.com/uuidjs/uuid/issues/750)) ([883b163](https://github.com/uuidjs/uuid/commit/883b163b9ab9d6655bfbd8a35e61a3c71674dfe1))
+- support rfc9562 v8 uuids ([#759](https://github.com/uuidjs/uuid/issues/759)) ([35a5342](https://github.com/uuidjs/uuid/commit/35a53428202657e402e6b4aa68f56c08194541bf))
+
+### Bug Fixes
+
+- revert "perf: remove superfluous call to toLowerCase ([#677](https://github.com/uuidjs/uuid/issues/677))" ([#738](https://github.com/uuidjs/uuid/issues/738)) ([e267b90](https://github.com/uuidjs/uuid/commit/e267b9073df1d0ce119ee53c0487fe76acb2be37))
```
ctavan, line 21: I don’t remember, but should this one have been release in the 9.x track?

ctavan (commented):
LGTM in general.

In the past we sometimes released new major releases as a prerelease and asked select folks who had been asking for breaking changes to test the pre release for their use case.

Do we want to do it here as well or just release and fix (if necessary)?

CHANGELOG.md
broofa: I don't recall either, but it doesn't affect functionality or APIs.  I'm fine with this in v10.

CHANGELOG.md
broofa: fixed

broofa:
> Do we want to do it [prerelease] here as well or just release and fix (if necessary)?

I thought about that.  Honestly, I don't really have the energy to deal with a prerelease.  I'm comfortable just pushing this out there and dealing with issues as-needed.

CHANGELOG.md
ctavan: Then point I wanted to make is if this perf change should be made available in a 9.x release so that folks who can’t easily upgrade (due to the breaking changes) can still get it?

broofa: I see... I misunderstood.

Yes, in principle, that patch should go out on v9.  In practice, that train has left the station.  I [briefly] opened a PR to patch the v9 release but doing so requires creating and maintaining a separate v9-maintenance branch because we already have v10 commits stacked on `main`.  While that's not a big deal, historically we've managed to avoid that sort of thing.  And given that v9.0.1 (with the regression) has been in prod for almost a year with no complaints, I'm see no reason to start now.

----

@pmccarren:  This is a small issue, but with some important context that you should have as a new maintainer.  Briefly...

Performance of `uuid` code is an issue-that's-never-actually-an-issue.  As `npm run test:benchmark` shows, UUIDs can be generated 100K's or M's of UUIDs per second.  That's orders of magnitude faster than any real-world application I'm aware of.  And yet there is a temptation to "tune" the code that's hard to resist.  I've been plenty guilty of this in the past, and we occasionally get PRs from well-intended contributors suggesting ways to make it faster.

I'm very leery of this sort of thing, however.  Such tuning typically relies on an innate knowledge of how JS runtimes work and usually introduces code that is, at best, "interesting" and makes things harder to maintain.  PR #434, which is what @ctavan and I are talking about here, is a great example of this.  In that case the benefit was enough we opted to take the change, but it's.... weird.   TL;DR: Calling `.toLowerCase()` improves V8 garbage collection and makes the code faster.

Go figure.

The problem with this is how non-sensical that is on the surface.  Not surprisingly, we forgot about why we made that change and ended up removing it later (#677).  Later still, we realized the mistake and I restored it (#738).  And because nobody actually cares about perf in the real world, I didn't bother pushing a new release.  I figured we'd just roll it out with whatever the next release was.... and here we are.

ctavan: Very well put! Let’s ship 10.0 asap then!

maon-fp:
Thanks a lot!

PR #763, opened 2024-06-10
feat!: Port to TypeScript, closes  #762

### LINKED ISSUE #762: Port codebase to TypeScript
Yet another issue for adding TS support to the project.  This time, however, we'll (hopefully) be doing a full port of the codebase to TS so we can use `tsc` to generate the type declaration files.

This will affect the build system, which may (read, "definitely") break backward compatibility.

See PR #763 for specific tasks.

broofa:
Switching the codebase to Typescript.

* [x] Move src/*.js -> src/*.ts
* [x] add type annotations throughout
* [x] Prefer `Uint8Array` for binary uuids throughout (BREAKING CHANGE)
* [x] Port unit tests to TS
* [x] Switch test runner from `jest` to `node:test` 
* [x] Switch to `tsc` for generating buidls
* [x] Get all checks passing

Other notes:
* **Changes to src/*.js files should be limited to...**
  * Adding type annotations
  * Switching to `Uint8Array` for binary UUIDs
* The `examples/node-jest` test passes, but I don't see it being run as part of CI or `npm test` anywhere.  I'm tempted to remove it. 🤔 
* Updated to `eslint@latest`, and switched config to ESM.
* Updated to `@wdio/*@latest`
* Switched `prettier` config to ESM
* Added VSCode configuration.  @ctavan, iirc you've objected to this in the past, but these are very helpful to me and to other vscode users that may want to contribute to this codebase.  Specifically:
  * `tasks.json` starts the `build:watch` and `test:watch` scripts automatically when you open the project
  * `extensions.json` identifies the plugins that devs should use to apply our `eslint` and `prettier` settings while editing code
* Sorted `package.json#scripts` alphabetically
  * Added `build:watch` and `test:watch` scripts, which I'm using in vscode
* Unit tests have been ported to TS, but should be functionally identical.
* Tests and CI are run using the `dist/esm` directory (unless they're explicitely testing other dist/* builds.
* `build.sh` now builds ESM and CJS versions of code using `tsc` rather than babel.
  * The top-level `dist/*.js` files can now be found in `dist/esm`, alongside the other builds (`esm-browser`, `cjs`, and `cjs-browser`)
* Switched to using `Uint8Array` for binary (byte) UUID data structures throughout.  The one exception is the return type for `md5` and `sha1` on node, which is still Buffer (but that's a subclass of Uint8Array these days anyways so... 🤷)


broofa:
@LinusU @pmccarren I would particularly appreciate a review of the TS types I've added, particularly with regard to API surface where I've used overload signatures to capture the different incantations that can be used.

@ctavan et al: Thoughts on release process?  I know we just pushed v10, so pushing a v11 this soon after feels a bit... sudden?  We could do a prerelease ("11.0.0-beta") and dist tag as `npm dist-tag add 11.0.0-beta beta`, I suppose.

broofa:
@joshkel As the most recent contributor to @types/uuid, I'd certainly appreciate any feedback you have on this.

broofa:
**Final call**.   Merging this by week's end unless someone objects.

@ctavan @LinusU @pmccarren @joshkel

tsconfig.esm.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "module": "ESNext",
+    "moduleResolution": "Node",
```
jaydenseric, line 5: For a published package (or even projects) it's not correct to use any `module` mode other than `nodenext`; that's the only mode that resolves things according to the latest Node.js ESM rules taking into account package `exports` field, imports having to have fully specified paths (including the file extension), etc.

broofa: Thanks for the feedback.

I've been writing TS for 4+ years now, 'have set up dozens of TS projects.  And yet whenever I have to set up a `tsconfig.json` file it always just feels a bit...

broofa: https://github.com/uuidjs/uuid/pull/778/files

PR #767, opened 2024-06-14
chore: add trend image

cwtuan:
Thank you for maintaining the "uuid" package. It has been incredibly useful for my projects, and I appreciate the effort you and the contributors have put into its development.

As an active user of this package, I would like to suggest a small enhancement to the README file that could benefit potential users. Specifically, I propose adding a downloads trend image.

New users often look for signs of a well-maintained and widely-used package. A positive download trend can be a persuasive factor. It will visually demonstrate the increasing adoption and popularity of the "uuid" package over recent years. Highlighting the growth of the package can encourage more contributions and usage from the community.

Thank you for review!

Here's a preview of what the updated README section would look like with the trend image:

image


broofa:
@cwtuan  Thanks for the PR, but our README is already pretty lengthy, and the places where users typically see it -  NPM and GitHub - already show `uuid` adoption #'s in the sidebar.

PR #775, opened 2024-07-10
fix(v7): seq_hi shift for byte 6

doomgirl:
A typo that fits just right in the rfc example blind spot.


broofa (changes requested):
Thanks for the contribution.  Good catch on the `seqHigh` bug.

I like that bit-flip test.  And the use of a BigInt form of UUID is clever, but does impact the readability.  I've requested some minor changes for readability and style as a result.

cc @pmccarren

broofa (approved):

PR #776, opened 2024-07-16
fix!: refactor v7 internal state and options logic, fixes #764

broofa:
Putting this PR up as an alternative to #768 that addresses #764 and a few other concerns.

* Written in Typescript (required now that #763 has landed)
* Decouple behavior of `v7()`-with-options from `v7()`-without-options.
* Encapsulate internal state handling and refactor to be more idiomatic
* Rigorous unit tests for internal state transitions
* Use 32-bit `seq` field rather than 31.

@robinpokorny My apologies for closing #768.  I wanted to merge it, but the switch to typescript and these other issues just made it easier to address in a new PR.  I'll do my best to credit your work here.

... and if you could help review this, I'd appreciate it. I feel pretty good about this refactor but there's enough code churn that an extra set of eyes would be very helpful.

Where #764 is concerned, the heart of this PR is [in this code](https://github.com/uuidjs/uuid/blob/894b6299c72c019c7b2a5897f41f60c8e390fb05/src/v7.ts#L20-L37), where the behavior very deliberately branches based on whether `options` are present.

> [!NOTE]
> * 'Marking this as a breaking change because of the changes to default values for `options#seq` and `options#msecs`.
> * The `v1()` code would probably benefit from a similar refactoring, if for nothing else than to make the APIs consistent.  I'll put up a PR for that shortly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,10 +338,10 @@ Create an RFC version 7 (random) UUID
 |  |  |
 | --- | --- |
 | [`options`] | `Object` with one or more of the following properties: |
-| [`options.msecs`] | RFC "timestamp" field (`Number` of milliseconds, unix epoch) |
+| [`options.msecs`] | RFC "timestamp" field (`Number` of milliseconds, unix epoch). Default = `Date.now()` |
 | [`options.random`] | `Array` of 16 random bytes (0-255) |
 | [`options.rng`] | Alternative to `options.random`, a `Function` that returns an `Array` of 16 random bytes (0-255) |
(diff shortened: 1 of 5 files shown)


broofa:
I have no idea why this got closed.   The `ts_port` branch has already been merged, which should've caused the base for this PR to change to `main`.  But I guess not?  So deleting `ts_port` closed this PR?

One of those "I guess I don't understand git/github after all moments".  'Will reopen in a new PR.  😕

PR #779, opened 2024-07-17
fix!: refactor v7 internal state and options logic, fixes #764

broofa:
[#776, reopened as a new PR.  No idea why it's including so many commits that are already on `main`]

Putting this PR up as an alternative to #768 that addresses #764 and a few other concerns.

* Written in Typescript (required now that #763 has landed)
* Decouple behavior of `v7()`-with-options from `v7()`-without-options.
* Encapsulate internal state handling and refactor to be more idiomatic
* Rigorous unit tests for internal state transitions
* Use 32-bit `seq` field rather than 31.

@robinpokorny My apologies for closing #768.  I wanted to merge it, but the switch to typescript and these other issues just made it easier to address in a new PR.  I'll do my best to credit your work here.

... and if you could help review this, I'd appreciate it. I feel pretty good about this refactor but there's enough code churn that an extra set of eyes would be very helpful.

Where #764 is concerned, the heart of this PR is [in this code](https://github.com/uuidjs/uuid/blob/894b6299c72c019c7b2a5897f41f60c8e390fb05/src/v7.ts#L20-L37), where the behavior very deliberately branches based on whether `options` are present.

> [!NOTE]
> * 'Marking this as a breaking change because of the changes to default values for `options#seq` and `options#msecs`.
> * The `v1()` code would probably benefit from a similar refactoring, if for nothing else than to make the APIs consistent.  I'll put up a PR for that shortly.


robinpokorny (approved):
I really like it now! I think the trade offs are well selected. Good job  🎉 

Maybe, in README we should write about the monotonicity and that providing any options means no guarantees of it. WDYT?

broofa:
@robinpokorny 

> Maybe, in README we should write about the monotonicity and that providing any options means no guarantees of it. WDYT?

I'll update the README once this PR and #780 land.
* 💯  on documenting the new options vs. no-options behavior.
* What did you have in mind re: monotonicity?  (The README is pretty long, so I generally try to avoid soap-boxing on topics that the RFC(s) cover.)

PR #782, opened 2024-07-21
chore: build script updates

broofa:
* ~~Add node@22 to test matrix~~
* Fix `runmd` build
* Change test:pack build dir to /tmp so it doesn't pollute source dir, and is not in-scope of project's package.json

> [!NOTE]
> Reverted the addition of `node@22` to test matrix because the `npm ci` command was failing with the mysterious ["Document exit handler not called" error](https://github.com/npm/statusboard/issues/395).  Given that ci.yml works for node@16,18 and 20, and only fails this way for node@22, and my limited time/interest in tracking this down, I'm going to chalk this up to a node regression and move on.


scripts/testpack.sh
```diff
@@ -4,20 +4,25 @@
 ROOT="$(pwd)/$(dirname "$0")/.."
 cd "$ROOT" || exit 1
 
-npm pack
+TEST_DIR=/tmp/test-pack
```
ctavan, line 7: Prefer mktemp + trap: https://unix.stackexchange.com/a/509189

ctavan (changes requested):

scripts/testpack.sh
broofa: Done.

scripts/testpack.sh
```diff
@@ -4,7 +4,8 @@
 ROOT="$(pwd)/$(dirname "$0")/.."
 cd "$ROOT" || exit 1
 
-TEST_DIR=/tmp/test-pack
+TEST_DIR=$(mktemp -d)
```
ctavan, line 7: Set up the trap right after this line so it takes effect if the script exits somewhere in the middle

broofa:
@ctavan Okay to merge?

scripts/testpack.sh
```diff
@@ -4,20 +4,26 @@
...
-npm pack
+TEST_DIR=$(mktemp -d)
 
-mkdir -p ../test-pack
+mkdir -p ${TEST_DIR}
 
-cp examples/node-commonjs/example.js ../test-pack/commonjs.js
-cp examples/node-esmodules/example.mjs ../test-pack/esmodules.mjs
-cp examples/node-esmodules/package.mjs ../test-pack/esmodules-package.mjs
+trap 'popd && rm -rf $TEST_DIR' EXIT
```
ctavan, line 11: I’d remove the popd from here… npm pack could fail and then you’d popd without having pushd’ed…

What we mostly care about with the trap is to no leave temp files.

If you want to overengineer, add a separate trap with just popd right after the pushd.

ctavan (approved):

PR #786, opened 2024-08-05
fix: remove v4 options default assignment preventing native.randomUUID from being used

molaeiali:
This removes the v4 options default assignment introduced during porting to ts (by mistake maybe?), during https://github.com/uuidjs/uuid/pull/763 - [PR diff](https://github.com/uuidjs/uuid/pull/763/files#diff-d7d9279b5eb8b7335eff565ca2e9a7e9ec8ff7d7de248197ce80481c118ea5de) - [commit](https://github.com/uuidjs/uuid/commit/1e0f9870db864ca93f7a69db0d468b5e1b7605e7#diff-d7d9279b5eb8b7335eff565ca2e9a7e9ec8ff7d7de248197ce80481c118ea5de) preventing `native.randomUUID` from being used

if `options` always have a value, we will never enter the if in line 11. `{}`  will be assigned to `options` in line 15 after the native check

it looks to me that the line 9 statement is added by mistake

### DIFF
--- a/src/test/v4.test.ts
+++ b/src/test/v4.test.ts
@@ -1,6 +1,7 @@
 import * as assert from 'assert';
-import test, { describe } from 'node:test';
+import test, { describe, mock } from 'node:test';
 import v4 from '../v4.js';
+import native from '../native.js';
 
 const randomBytesFixture = Uint8Array.of(
   0x10,
@@ -46,6 +47,26 @@ describe('v4', () => {
     const id2 = v4();
 
     assert.ok(id1 !== id2);
+  });
+
+  test('should uses native randomUUID() if no option is passed', () => {
+    mock.method(native, 'randomUUID');
+
+    assert.equal((native.randomUUID as any).mock.callCount(), 0);
+    v4();
+    assert.equal((native.randomUUID as any).mock.callCount(), 1);
+
+    mock.restoreAll();
+  });
+
+  test('should not use native randomUUID() if an option is passed', () => {
+    mock.method(native, 'randomUUID');
+
+    assert.equal((native.randomUUID as any).mock.callCount(), 0);
+    v4({});
+    assert.equal((native.randomUUID as any).mock.callCount(), 0);
+
+    mock.restoreAll();
   });
 
   test('explicit options.random produces expected result', () => {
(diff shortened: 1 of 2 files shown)


molaeiali:
@broofa I've added some tests that I think are cleaner than adding a variable inside the main code and counting that!

But there's a minor issue with these tests, as you can see here:
```
 if (native.randomUUID && !buf && !options) {
```
the first condition is for `native.randomUUID` to exist, I think it doesn't exist in node<14 so if tests are run with node<14 the first test fails, is that an issue?

src/test/v4.test.ts
```diff
@@ -48,6 +49,26 @@ describe('v4', () => {
...
   });
 
+  test('should uses native randomUUID() if no option is passed', () => {
+    mock.method(native, 'randomUUID');
+
+    assert.equal((native.randomUUID as any).mock.callCount(), 0);
+    v4();
+    assert.equal((native.randomUUID as any).mock.callCount(), 1);
+
+    mock.restoreAll();
```
broofa, line 59: Cool! I wasn't aware of the `mock` feature of Node's test runner until now.  Excellent change!

BTW, it looks like `mock.method()` return value is the mocked function with the proper TS "`Mock`" type, so we can capture that to avoid having to cast to the `any` type (which is always a bit of a code smell).

```suggestion
    const mockedV4 = mock.method(native, 'randomUUID');

    assert.equal(mockedV4.mock.callCount(), 0);
    v4();
    assert.equal(mockedV4.mock.callCount(), 1);

    mock.restoreAll();
```

src/test/v4.test.ts
```diff
@@ -48,6 +49,26 @@ describe('v4', () => {
...
+  });
+
+  test('should not use native randomUUID() if an option is passed', () => {
+    mock.method(native, 'randomUUID');
+
+    assert.equal((native.randomUUID as any).mock.callCount(), 0);
+    v4({});
+    assert.equal((native.randomUUID as any).mock.callCount(), 0);
+
+    mock.restoreAll();
```
broofa, line 69: Same change as above.

broofa (changes requested):
One last change and I think we'll be ready to merge.

broofa (approved):
Approving and merging.  I'll make the suggested type changes in a separate PR.

Thanks for the contribution!

molaeiali:
Sorry I was a bit busy, thanks for responding and being open to PRs

PR #787, opened 2024-08-06
README says uuid@11 has been published (including types), but latest version on NPM is uuid@10 (doesn't include types)

aoor9:
Even though this project is now built with typescript, I get this error when I use it on my node app:
```
Could not find a declaration file for module 'uuid'. './app/node_modules/uuid/dist/esm-browser/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/uuid` if it exists or add a new declaration (.d.ts) file containing `declare module 'uuid';`ts(7016)
```
This fix exports types declaration from `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,11 +29,19 @@
       "node": {
         "module": "./dist/esm/index.js",
         "require": "./dist/cjs/index.js",
-        "import": "./wrapper.mjs"
+        "import": "./wrapper.mjs",
+        "types": {
+          "import": "./dist/esm/index.d.ts",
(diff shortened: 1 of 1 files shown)


broofa:
Sorry, `uuid@11` has not actually been published yet.  It's still a work in progress, and `uuid@10` (the `latest` version) doesn't have types included with it yet.   So, for now, you still need to `npm install @types/uuid` despite what the README says.

Sorry about that!  I expect to have v11 published in the next 1-2 weeks, and you should be good to go then.

PR #789, opened 2024-08-08
fix: remaining issues for v11 release

broofa:
Tracking and fixing a few remaining items before pushing out v11 pre-release:
* [x] Fixed build.sh code that moves "[name]-browser.\*" files to "[name].\*"
* [x] Fixed examples/node-esmodules and examples/browser-esmodules code for testing package.json import
* [x] Removed code from v35.test.ts that depended on "md5-browser.js" and "sha-browser.js", which are no longer present in `dist`.
* [x] Fix BrowserStack builds

@ctavan I know you're busy with other stuff these days, but I could use some help with the BrowserStack issue.  The `browser` workflow in **Checks** runs and passes, but I don't think it's actually doing anything.  [The log shows all tests as being skipped](https://github.com/uuidjs/uuid/actions/runs/10304139601/job/28521841099?pr=789#step:5:179).  Nor is there any meaningful logs in the BS dashboard.  For example that run links to [this empty output](https://observability.browserstack.com/builds/tmxyzv3qlrhrzbhp0kodstdaxkc61ptnvr0foxzj).

'Been this way for awhile.   FWIW, I see the same behavior when running the tests locally using [this config](https://gist.github.com/broofa/1de6d1ef1254bedeaa68d433ca1d9174) with `wdio -l debug run ./wdio.conf.js`.

Anyhow... I'll keep poking at this, but if you have any ideas for what might be going on, I'd be interested in hearing them.


broofa:
@ctavan 'Finally figured out the BrowserStack issue.  Apparently the`test(...)` method is no longer defined in the context of browser.spec.js.  How...?  Why...?  Dunno.  Probably happened when I updated `devDependencies` at some point.  Switching to `it(...)` fixes the problem.  'Would've been nice if there'd been some sort of error message in the logs but there wasn't.  Just "test skipped".  🤦 

Anyhow... that CI check is working again. Note, too, that I've removed the legacy `capabilities` (chrome@52, FF@51, Edge@15, Saf@11).  Once this goes in, we'll only be testing `browserVersion: 'latest'`.

PR #791, opened 2024-09-05
chore(release): 11.0.0-0

broofa:


broofa:
Browserstack tests are flaky.  Had to run the `browser` workflow three times before it passed, but it did pass so I'm merging.

PR #796, opened 2024-10-23
fix: remove obsolete code for removed IE11 support

Rotzbua:
IE11 / `msCrypto` was removed with release v9.


broofa:
Good catch.  Thanks for the PR!

PR #510, opened 2020-08-14
test: install WebdriverIO packages as optional dev dependencies

ctavan:

PR #802, opened 2024-10-25
chore: fix optionalDevDependencies regression

broofa:
Fixing a regression with `optionalDevDependencies`.

@IgorTodorovskiIBM please comment on this to let me know if issue #509 is still relevant.   The `optional-dev-dependency` util is a bit of a footgun because of how it mutates `package.json` unexpectedly, and how easy it is to forget that it's part of our dev environment.  That, and the fact this regression has gone unnoticed for four months, has me wondering if we could revert #510 to simplify our build setup... which would be really nice.

I'm fixing this for now, but if I don't hear anything to the contrary  I expect I'll do that reversion the next time this issue pops up.


IgorTodorovskiIBM:
Actually, this PR would help as our z/OS uuid builds continue to fail because wdio is added as a devDependencies. Moving it out of devDependencies to only in optionalDevDependencies would help.

ctavan:
Thanks a lot for all the recent cleanups @broofa!

Just noticed one thing: are we able to submit PRs without approval from a second maintainer now?

Issue #820, opened 2024-10-28
[BUG] (v11) Unable to start Node.js app after build removed `wrapper.mjs`


peterhirn:
### Before you begin...

- [X] I have searched the existing issues
- [X] I am not using version 13.x of node  (if so, please upgrade)

### Description of the problem

Unclear if this is a bug here or in [vinxi](https://github.com/nksaraf/vinxi)

### Recipe for reproducing

`pnpm create solid@latest -s solid-start-uuid-v11`

- template: basic
- typescript: Yes

```
cd solid-start-uuid-v11
pnpm add uuid
```

Edit `src/entry-server.tsx`

```ts
import { v7 } from "uuid";
console.log(v7());
```

```
pnpm build
pnpm start
```

```
 ERROR  Cannot find module '/.../solid-start-uuid-v11/.output/server/node_modules/uuid/wrapper.mjs' imported from /.../solid-start-uuid-v11/.output/server/index.mjs              8:23:43 AM

  at finalizeResolution (node:internal/modules/esm/resolve:275:11)
  at moduleResolve (node:internal/modules/esm/resolve:932:10)
  at defaultResolve (node:internal/modules/esm/resolve:1056:11)
  at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:650:12)
  at #cachedDefaultResolve (node:internal/modules/esm/loader:599:25)
  at ModuleLoader.resolve (node:internal/modules/esm/loader:582:38)
  at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:241:38)
  at ModuleJob._link (node:internal/modules/esm/module_job:132:49)

```

### Additional information

`pnpm dev` works as expected.

### Environment

Node: v23.1.0

## Workaround

`pnpm patch uuid`

```patch
diff --git a/package.json b/package.json
index 12ea33f0527be28f3d263ef32b5a189882b46455..8475de8aea93583640cbb7ebfa966ed262e27635 100644
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
       "node": {
         "module": "./dist/esm/index.js",
         "require": "./dist/cjs/index.js",
-        "import": "./wrapper.mjs"
+        "import": "./dist/esm/index.js"
       },
       "browser": {
         "import": "./dist/esm-browser/index.js",
```

broofa:
Thanks for the report.  I'm able to reproduce the issue.  Looking into it.

broofa:
[...]

PR #822, opened 2024-10-28
fix: remove wrapper.mjs

broofa:
@ctavan @TrySound :  Do we still need the wrapper.mjs file in light of the `uuid@v11` refactoring / changes to `node` since #423 went in?

#820 Isn't an issue with `uuid` per se, but it does speak to the additional code-complexity that comes with the wrapper approach.  It'd be nice to get rid of it.

I've scanned #423 but my grasp of the issues there is a bit dated.  I need to refresh my memory.  In the meantime I'm unsure how relevant this still is.


broofa:
Looking into this further I'm 99% sure we can kill wrapper.mjs.  It was intended as a way to bridge node's experimental ESM support to the CJS code base, which was the preferred implementation at the time.   That's not really a relevant concern any longer... not for me, at least. 😁   In fact, in `uuid@10` the file wrapper.mjs was importing (dist/index.js) was CJS, not ESM, but it now points to dist/esm/index.js which is, at best, a benign bug that makes wrapper.mjs synonomous with dist/esm/index.js in `uuid@v11.0.1` currently.

So... in the interests of mitigating the # of users impacted by this change I'm going to go ahead and push a new patch-release.  For anyone impacted by this... uh... "I'm sorry"... and please comment here or open a new issue where we can discuss how / why we need to continue supporting your use case.

PR #830, opened 2024-10-31
chore: update to actions/checkout@v4

broofa:
Fixes this warning: 
CleanShot 2024-10-31 at 07 02 23@2x

### DIFF
--- a/.github/workflows/browser.yml
+++ b/.github/workflows/browser.yml
@@ -15,7 +15,7 @@ jobs:
     timeout-minutes: 30
 
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/browser.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     timeout-minutes: 30
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 10
```
broofa, line 20: @ctavan : Do you recall why we specify `fetch-depth` here?

PR #831, opened 2024-11-01
fix: apply stricter typing to the v* signatures

robbtraister:
The `v4` function is guaranteed to return a string unless the `buf` argument is provided. By modifying the signature to denote this parameter as required for the `Uint8Array` case, we get the benefit of a stricter return type of `string` when simply calling `v4()`.

### DIFF
--- a/src/v1.ts
+++ b/src/v1.ts
@@ -25,7 +25,7 @@ type V1State = {
 const _state: V1State = {};
 
 function v1(options?: Version1Options, buf?: undefined, offset?: number): string;
-function v1(options?: Version1Options, buf?: Uint8Array, offset?: number): Uint8Array;
+function v1(options: Version1Options | undefined, buf: Uint8Array, offset?: number): Uint8Array;
 function v1(options?: Version1Options, buf?: Uint8Array, offset?: number): UUIDTypes {
   let bytes: Uint8Array;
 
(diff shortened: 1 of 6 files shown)


broofa (changes requested):
Nice!  Thanks for the contribution.

Can you do me a favor and make update the other methods that take an optional `buf` argument as well? (v1.ts, v3.ts, v5.ts, and v7.ts)

broofa:
@robbtraister Thanks for the update.  This looks good.

FYI, I added [this commit](https://github.com/uuidjs/uuid/pull/831/commits/2e09800e3000d01789ddfd688d0c21578b4b2ad7) after realizing the overload types for v3 and v5 aren't really necessary.  (Right?  If I got that wrong, please let me know.)

broofa (approved):

robbtraister:
> @robbtraister Thanks for the update. This looks good.
> 
> FYI, I added [this commit](https://github.com/uuidjs/uuid/pull/831/commits/2e09800e3000d01789ddfd688d0c21578b4b2ad7) after realizing the overload types for v3 and v5 aren't really necessary. (Right? If I got that wrong, please let me know.)

Calling `v3("value", "namespace")` should always return `string`, but removing the overloads types it as `string | Uint8Array`.

Having said that, I only ever use `v4`, so 🤷

broofa:
> Calling v3("value", "namespace") should always return string, but removing the overloads types it as string | Uint8Array.

Ah, right... thanks for catching that!

PR #833, opened 2024-11-03
fix: exporting types

### LINKED ISSUE #832: Export options types
### Feature description

I was previously using the options type provided in the DefinitelyTyped package (i.e. `V7Options`) to type a function that passed through options to the UUID constructor. I would love if those options types were exported by UUID v11

### Additional information

I can work around the issue by declaring a type `type V7Options = Parameters<typeof v7>[0]`, but it's a little clunky compared to having access to the type directly

skycoop:
This exports the types from `_types.ts` in the package's entrypoint. Resolves #832

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,10 @@
+export type {
+  UUIDTypes,
+  Version1Options,
+  Version4Options,
+  Version6Options,
+  Version7Options,
+} from './_types.js';
 export { default as MAX } from './max.js';
 export { default as NIL } from './nil.js';
 export { default as parse } from './parse.js';


src/index.ts
```diff
@@ -1,3 +1,10 @@
+export type {
+  UUIDTypes,
+  Version1Options,
+  Version4Options,
+  Version6Options,
+  Version7Options,
+} from './_types.js';
```
broofa, line 7:
```suggestion
export type * from './_types.js';
```

This appears to work and will be a bit more future proof as types get added / removed. (please test for your use case, however)

Note:  "_types" name implies these are _private_ types, which is no longer the case.   Please rename src/_types.ts to src/types.ts, and updated references to it accordingly.

broofa (changes requested):

broofa (approved):

src/index.ts
bobvandevijver: @broofa This change breaks compatibility with older Typescript versions, as export type * is only allowed since Typescript v5. See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#support-for-export-type-.

Maybe revert back to the explicit export?

PR #836, opened 2024-11-04
chore: replace optional-dev-dependencies with iodd

broofa:
Replaces the defunct `optional-dev-dependencies` utility with the newly minted `iodd` script.

`iodd` is little more than an alias for `npm install --no-save`, passing it the packages defined in `package.json#optionalDevDependencies`.  In fact, we could just hard-code a `package.json#script` that runs the same command.  'Might even be simpler, but `iodd` has the advantage of exiting gracefully if the packages fail to install, which is nice.

See also: https://github.com/bcoe/optional-dev-dependency/issues/23


broofa:
@ctavan: This seems to work (all checks passing), but I'll wait for your review on this since this was your code originally.

ctavan (approved):
LGTM, I think this is a good workaround!

PR #837, opened 2024-11-11
fix: compatibility with Typescript V4

bobvandevijver:
With https://github.com/uuidjs/uuid/pull/833 the type export was added, but the `export type *` syntax is only available since Typescript V5. Using direct imports should resolve the issue.

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#support-for-export-type- for the documentation on this.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,10 @@
-export type * from './types.js';
+export type {
+  UUIDTypes,
+  Version1Options,
+  Version4Options,
+  Version6Options,
+  Version7Options,
+} from './types.js';
 export { default as MAX } from './max.js';
 export { default as NIL } from './nil.js';
 export { default as parse } from './parse.js';


broofa:
Thanks for the contribution.  

FWIW, I specifically [requested the wildcard export](https://github.com/uuidjs/uuid/pull/833#discussion_r1827885116) to make this code more future-proof.  I am admittedly reluctant to go back on that.

This prompted me to do a bit of digging to figure out what our TS support policy should be since that hasn't really been discussed yet.  Based on [this comment](https://github.com/microsoft/TypeScript/issues/49088#issuecomment-2468723715), it sounds like there's good precedent for adopting a policy of supporting TS versions less than two years old, only.   This also aligns roughly with our long-standing `node` support policy (LTS versions + one prior).

As it turns out, TS 4.9 - the last 4.X version released - will fall below that bar at the end of this week.  I promise, I did not plan this. 😄 

Given the above, I don't see a need for this change.

broofa:
TS4.9.5 appears to work just fine with `uuid@11.0.3`, both on the command line and VSCode.  Here's the test I used....

[uuid-ts-types-test.zip](https://github.com/user-attachments/files/17719589/uuid-ts-types-test.zip)

Given that and that uuid@11 has only been out for three weeks, and TS@4 falling off the support matrix for other similar projects (Definitely typed, typescript-eslint)... I'm okay leaving things as they stand.

bobvandevijver:
@broofa That would be because you have explicitly disabled the check in your test with `skipLibCheck: true`. If you remove that line you will see the error directly.

PR #840, opened 2024-11-21
fix: exclude tests from published package

kyletsang:
Current versions of v11 are including the tests in the published package.

This PR fixes that by preventing transpilation of the test folder in the root tsconfig and uses a test-specific tsconfig which does transpile test files.  The `tsconfig.test.json` only targets esm since the original test command uses that folder only.  Declarations are not emitted because they are not used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "pretest:benchmark": "npm run build",
     "pretest:browser": "./scripts/iodd && npm run build && npm-run-all --parallel examples:browser:**",
     "pretest:node": "npm run build",
-    "pretest": "npm run build",
+    "pretest": "tsc -p tsconfig.test.json",
     "prettier:check": "prettier --check .",
     "prettier:fix": "prettier --write .",
     "release": "standard-version --no-verify",
--- a/package.json
+++ b/package.json
@@ -44,11 +44,8 @@
     "./dist/cjs/index.js": "./dist/cjs-browser/index.js"
   },
   "files": [
-    "CHANGELOG.md",
-    "CONTRIBUTING.md",
-    "LICENSE.md",
-    "README.md",
-    "dist"
+    "dist",
+    "!dist/**/test"
   ],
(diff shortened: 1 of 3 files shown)


broofa:
Hey, thanks for catching this.  I agree removing the test files probably makes sense.  (That said I vaguely recall a conversation with _someone_ a few years back making the case for including them...?  But I don't recall when / where and  I can't make a good case for keeping them.)

Instead of having a separate build pipeline for the test files - which has a whiff of code smell to it let's add an [`.npmignore`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files) file with the necessary incantation(s) for excluding the test dir(s).

kyletsang:
Hey @broofa, I looked over this today with the `npmignore` and ran into this from the [docs](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files):

> At the root of your package it will not override the "files" field, but in subdirectories it will.

I wasn't able to get the `npmignore` to exclude the test dir.  Another approach would be to switch out the `"dist"` entry in `files` with a glob like `"dist/{cjs,cjs-browser,esm,esm-browser}/*.{js,d.ts}"`.  

Thoughts?

broofa:
Oof, sorry.  'Didn't realize that .npmignore was so "ineffective".

Please do the following:
1.  Revert the current changes.
2. Change `package.json#files` to be as follows:

```jsonc
  "files": [
    "dist",
    "!dist/**/test"
  ]
```

And, yes, remove the *.md entries that are currently there.  `npm` includes README and LICENSE by default, and the CHANGELOG and CONTRIBUTING docs don't need to be in the published package.

BTW, in case you're not aware `npm pack --dry-run` will show you which files will be published.  I've tested the change above and "it works for me", but please verify.

kyletsang:
Confirmed on my end, everything is as expected.  Thanks!

broofa (approved):

PR #844, opened 2024-12-04
docs: Ensure link to getrandomvalues-not-supported is maintained

dafrito:
When using on React Native, there is a link shown, https://github.com/uuidjs/uuid#getrandomvalues-not-supported, if the polyfill react-native-get-random-values is not installed. However, this link's anchor is not working because the title in the README has changed.

The React Native info was removed in b08c122248950db0df371c79638e0c5ffe3d6726 and re-added in ae8f38657bca0ee053bf29c88c006b1ea05af1b5, but the heading was changed so this link was broken.

This patch re-adds that anchor so the link points the user to proper spot in the README.

### DIFF
--- a/README_js.md
+++ b/README_js.md
@@ -498,7 +498,12 @@ Prior to `uuid@11`, it was possible for `options` state to interfere with the in
 
 ## Known issues
 
-### React Native / Expo
+<!-- This header is referenced as an anchor in src/rng-browser.ts -->
+### "getRandomValues() not supported"
+
+This error occurs in environments where the standard [`crypto.getRandomValues()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) API is not supported. This issue can be resolved by adding an appropriate polyfill:
+
+#### React Native / Expo
 
 1. Install [`react-native-get-random-values`](https://github.com/LinusU/react-native-get-random-values#readme)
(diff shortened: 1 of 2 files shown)


README_js.md
```diff
@@ -500,6 +500,7 @@ Prior to `uuid@11`, it was possible for `options` state to interfere with the in
 
 ### React Native / Expo
 
+<a name="getrandomvalues-not-supported"></a>
```
broofa, line 503: Good catch, but this isn't the markdown that got removed.  Let's this and restore these lines above this section (where they were previously), please:

https://github.com/uuidjs/uuid/blob/bc46e198ab06311a9d82d3c9c6222062dd27f760/README_js.md?plain=1#L385-L387,

broofa, line 503: Good catch, but this isn't the markdown that got removed.  Let's this and restore these lines above this section (where they were previously), please:

https://github.com/uuidjs/uuid/blob/bc46e198ab06311a9d82d3c9c6222062dd27f760/README_js.md?plain=1#L385-L387,

broofa (changes requested):

README_js.md
dafrito: Fixed by re-adding the header title as suggested. I shrunk the "React Native/Expo" header so it appears as a subsection. I also added a HTML comment to document the reference of this section.

broofa:
Thanks for catching this.

818Nawaf (commented):
موافق

PR #845, opened 2024-12-24
fix: v4 UUID generation with insufficient random bytes should match e…

gaoyia:
```TS

const v4options = {
  random: Uint8Array.of(
    0x10,
    0x91,
    0x56,
    0xbe,
    0xc4,
    0xfb,
    0xc1,
    0xea,
    0x71,
    0xb4,
    0xef,
    0xe1,
    0x67,
    0x1c // Missing 2 item
  ),
};
uuidv4(v4options); // ⇨ '109156be-c4fb-41ea-b1b4-efe1671cundefinedundefined'

### DIFF
--- a/src/test/v4.test.ts
+++ b/src/test/v4.test.ts
@@ -2,6 +2,9 @@ import * as assert from 'assert';
 import test, { describe } from 'node:test';
 import native from '../native.js';
 import v4 from '../v4.js';
+import parse from '../parse.js';
+import version from '../version.js';
+import { unsafeStringify } from '../stringify.js';
 
 const randomBytesFixture = Uint8Array.of(
   0x10,
@@ -21,7 +24,7 @@ const randomBytesFixture = Uint8Array.of(
   0x58,
   0x36
 );
-
+const randomBytesFixtureLess = Uint8Array.of(16);
 const expectedBytes = Uint8Array.of(
   16,
   145,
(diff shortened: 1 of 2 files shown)


gaoyia:
maybe we shoud 

- [x] use random value
- [ ] throw a Error
- [ ]  pad with zeros： uuidv4(v4options); // ⇨ '109156be-c4fb-41ea-b1b4-efe1671c0000'

broofa:
Sorry, I missed your last comment.  I've updated the PR as follows:
* `throw` if `options.random` or `options.rng()` resultsin insufficient entropy (too-short array)
* `throw` if byte indexes over/underflow `buffer` range
* Apply chagnes to `v1()` and `v7()` as well

Note: I removed your code that silently switched to `rng()` if the user-supplied `options.random` or `options.rng()` produced a too-short array.  Rationale: If a user is supplying those options, they're likely not expecting a different RNG source to be used.  Better to be explicit about the problem rather than make assumptions about the solution.

broofa:
Note: CI / lint test is failing for reasons unrelated to this PR.  (`runmd` and `prettier` disagree on how the markdown should be formatted 😠 )

@ctavan : Okay to push this out as a patch release?   I only hesitate because this code may throw where it didn't previously.  While that's not strictly a breaking API change, I could see it catching some users unawares.  (But those users would've been getting pretty flagrantly bad UUIDs previously)

I'm surprised nobody has complained about these issues previously.  (FWIW, I vaguely remember encountering this problem years ago and thinking at the time that "undefined" appearing in a UUID string would throw for other reasons.   But that may have been before `buf` support was fleshed out?  Or before the v8 or v10 refactors?  I forget.)

broofa:
Thanks for the help!

PR #849, opened 2024-12-29
chore: add shellcheck linting

codecriticsceo:
PR adds the `shellcheck` linter to check the shell scripts, and fixes the found issues. Nothing major, although the existing scripts might not work if there are spaces or asterisks in the development repo path. If you would like changes, please let me know.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,8 @@ jobs:
       - run: npm ci
       - run: npm run lint
       - run: npm run docs:diff
+      - name: Run Shellcheck
+        run: find . -name '*.sh' -print0 | xargs -0 shellcheck -x
 
   test:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       - run: npm run lint
       - run: npm run docs:diff
       - name: Run Shellcheck
-        run: find . -name '*.sh' -print0 | xargs -0 shellcheck -x
+        run: find . -type f -name "*.sh" -exec shellcheck {} +
 
   test:
     runs-on: ubuntu-latest
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -26,24 +26,24 @@ for MODULE_TYPE in esm cjs; do
   DIST_DIR="$DIR/${MODULE_TYPE}"
   BROWSER_DIR="$DIR/${MODULE_TYPE}-browser"
 
-  tsc -p tsconfig.${MODULE_TYPE}.json
+  tsc -p "tsconfig.${MODULE_TYPE}.json"
 
   # Clone files for browser builds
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -17,6 +17,8 @@ jobs:
       - run: npm ci
       - run: npm run lint
       - run: npm run docs:diff
+      - name: Run Shellcheck
+        run: find . -name '*.sh' -print0 | xargs -0 shellcheck -x
```
broofa, line 21: Why use `find`?   Is the `-x` option necessary?  (What's it do?)

```suggestion
      - run: shellcheck **/*.sh
```

For that matter, why not use [the shellcheck GH action](https://github.com/marketplace/actions/shellcheck)?

broofa:
Noting for posterity that `shellcheck -f diff **/*.sh | git apply` can be used to auto-fix `shellcheck` issues.

.github/workflows/ci.yml
codecriticsceo: The reason I did not use it was because I don't think that's an official Shellcheck project action. Here is the page on shellcheck: https://www.shellcheck.net/wiki/GitHub-Actions. Would be happy to switch, if you want though.

broofa: Ah, okay.  That seems simple enough.  Maybe just change this to the recommended incantation...
```yaml
    - run: shellcheck **/*.sh
```

I don't think the `find` command is needed.  Nor the `-x` option (what does that even do?)

codecriticsceo: The -x option follows included shellscripts and checks them as well. The find command is so that shellscripts in varying directory levels (including root) will all be found.

codecriticsceo: I can test with the action, but in bash, `echo **/*.sh` appears not to find shellscripts in the immediate directory, only a subdirectory

codecriticsceo: Update tested: https://github.com/codecriticsceo/uuid/actions/runs/12554064345/job/35002278202. I added two bad scripts to a throw-away branch, one in root and one two directories down. `shellcheck **/*.sh` did not find them but the `find` command did. I am going to try updating the shellcheck wiki.

codecriticsceo: Updated the PR with the exact command from the docs (find without -x or xargs)

broofa:
@codecriticsceo Hey, I appreciate the contribution, but I'm going to pass on this (for now).  I'm open to revisiting this if we encounter actual problems that `shellcheck` would have caught, but that hasn't happened yet.

Meanwhile, I'm not a fan of the DX friction I foresee this adding.  Having a CI-blocking check that isn't enforced earlier in the dev experience is invariably annoying.  And if the check fails for spurious reasons like "missing totally-optional  quotes"... well... that sort of thing gets old quick.

Thanks for the contribution, though, and I appreciate you putting `shellcheck` on my radar.   I'll definitely be keeping it in mind for other projects.

PR #859, opened 2025-01-09
build: use `npm pack` tarball for local examples and testing instead of .local

broofa:
@ctavan,  I don't expect this to be controversial but it'd be nice to get your sign off here since the ".local" directory setup was initially your idea.

Briefly, this does the following:
1. Have the build script create an `npm pack` tarball (in .build/uuid.tgz) that can be installed locally
2. Have all examples use that instead of the ".local" dir
3. Remove the .local dir

> [!NOTE]
> `npm install` in the examples/* dirs was failing on GH test runners because `npm pack`'s output is platform specific.   This tends to cause `npm` to complain when it compares the checksum it sees on one platform (GH runner) to the `integrity` checksum it caches in package-lock.json on another platform (like our laptops).  The workaround was to add the `--no-package-lock` when installing in examples/* dirs.  This isn't a big deal since the only thing being installed is the `uuid` tarball.   And it has the nice side-effect of removing the need for package-lock.json files in those dirs.


broofa:
Merging, as I need this for #860

PR #863, opened 2025-01-11
docs: made a link replacement in shields Browser

Radovenchyk:


broofa:
As I said previously, you'll need to run `npm run docs` to update README.md file with this change.  This is why the CI/lint check fails.  (I know this is a hassle, but it's how this project works.)

broofa:
Fixed in #864.  Thanks for catching this!

PR #870, opened 2025-04-20
refactor: improve imports for the Node assert and test modules

tomherni:
Just some test-related improvements I thought would be nice to contribute.

Node's assert module:
* Use protocol import (prefixed with `node:`).
* Use [strict assertion mode](https://nodejs.org/api/assert.html#strict-assertion-mode) (import from `node:assert/strict`).

Node's test runner:
* Import `test` as a named import, rather than a default import.


broofa:
TIL: `node/strict` is a thing.  This is a nice little level-up on code "tightness".  Thanks!

PR #878, opened 2025-06-17
Add tests for NodeJS 24

BrodaNoel:
Adding tests for NodeJS 24.

I've read in another issue in this repo that NodeJS 16 support was going to end in ~2 months ago.
We may have to consider removing it from tests or any other part it needs changes


BrodaNoel:
BTW, be careful, because seems like tests on 22.x are not running. Check the logs below.

Screen Shot 2025-06-16 at 22 14 21

PR #882, opened 2025-06-20
feat!: remove CommonJS support

### LINKED ISSUE #881: Drop CommonJS Support
[@ctavan I know we've discussed this previously, but I recently had to rework the [logic for detecting module type in npmgraph](https://github.com/npmgraph/npmgraph/blob/main/components/GraphPane/colorizers/ModuleTypeColorizer.tsx#L51-L125) and it reminded me of just how ridiculous this ESM+CJS world is.  It's just cemented in my mind that the dual-package "solution" is long-term horrible and that package owners should be more proactive about addressing the problem.]

CommonJS is an inferior module format and this "dual module" world the JS ecosystem finds itself in is awful.  "Dual module" packages...

* ... have more complicated build tooling
* ... have more complicated testing and CI
* ... have larger package footprints
* ... are more complicated to process by downstream tools

... all of which results in more pain for maintainers and/or users.

Sadly, nobody seems able to do anything about this.  Ideally the NPM and Node organizations would be actively advocating for CJS to be deprecated, but they seem ~~content~~ handcuffed(?) to limiting their involvement to [baby steps](https://github.com/nodejs/node/issues/33954) that do little to address the underlying problem.  Thus, I see the burden as falling on project maintainers.

I know there's no real demand for this from `uuid` users, but that's to be expected.  Nobody ever complains about a project supporting _too many_ platforms.  But we're at at a good place to be making this move.  RFC9562 is fully supported, we have TS support, and overall `uuid` is mature and stable.

I expect most users won't notice.  They'll be using toolchains that deal with modules in either format.   Those that care can pin their dependency to `uuid@11`.

(Note: My "users won't care" opinion is informed by my experience making `mime` ESM-only.  Dropping CJS support for that library took place with barely a peep from users.)

Refs:
https://github.com/sindresorhus/meta/discussions/15
https://github.com/nodejs/node/issues/33954

broofa:
PR showing what happens if we drop CommonJS support.

Briefly:
* Update README verbiage.  (Note: fixes minor bug in runmd setup, causing UUIDs to change)
* Switch examples/browser-rollup config from CJS -> ESM
* Drop CJS test from examples/browser-webpack
* Drop examples/node-commonjs
* Switch examples/node-jest from CJS -> ESM.  (Requires upgrading to jest@30)
* Remove CJS logic from build.sh
  * Also change  from four dist subdirs - "dist/[ esm | esm-browser | cjs | cjs-browser ]" - to just "dist" and "dist-browser"
* (cleanup) Only bundle one set of TS types (in the "dist" dir)
  * `types` field to explicitly point to "dist/index.d.ts"
* Simplify package.json fields (fwiw, this is motivated by how @sindresorhus appears to be structuring package.json for his 100's of popular ESM-only packages.)
  * Remove `main` field (only used for CJS entry points)
  * Remove `module` field (it was never a standard and has been replaced by "exports")
  * In `exports`, remove `import` and `require` instead specify `default` (esm) and `browser` (esm-browser)
* Replace the 4 tsconfig.* files we'd been using to handle the multiple TS configs for ESM & CJS with a single tsconfig.json file for the ESM config.

The net effect of the above:
* Source files deleted: 8
* Source lines of code deleted: 271
* Installed package file size: 788K -> 312K
* Installed package file count: 197 -> 77

Closes #881


broofa:
@43081j 12.x will go out at the end of next month to allow for updating to TS 5.2 at the same time..  See #887

PR #894, opened 2025-08-11
fix: improve v4() performance

### LINKED ISSUE #892: [BUG] v4() is far slower than randomUUID
### Before you begin...

- [x] I have searched the existing issues
- [x] My issue title is descriptive and specific to the problem (i.e. search-engine friendly)
- [x] I understand what an "MRE" is, and why providing one is important.  (If not, [read this](https://stackoverflow.com/help/minimal-reproducible-example))

### MRE and Description

```bash
Given the following file (index.js):


const { v4 } = require('uuid');
const { randomUUID } = require('crypto');

const iterations = 1e8;
{
  const start = performance.now();
  for (let i = 0; i < iterations; ++i) {
    v4();
  }
  // Outputs
  console.log(`v4() took ${(performance.now() - start).toFixed(0)}ms`);
}

{
  const start = performance.now();
  for (let i = 0; i < iterations; ++i) {
    randomUUID();
  }
  console.log(`randomUUID() took ${(performance.now() - start).toFixed(0)}ms`);
}


And `package.json`:


{
  "name": "slow-native-uuid",
  "version": "1.0.0",
  "description": "Demo of unexpectedly slow uuidv4 performance",
  "main": "index.js",
  "author": "Patrick Nappa",
  "license": "MIT",
  "dependencies": {
    "uuid": "^11.1.0"
  }
}


The performance of v4() is significantly slower than `randomUUID()`, despite `v4()` using `randomUUID()`.

Output:


v4() took 7270ms
randomUUID() took 2718ms


Here's a snippet of what `uuid/dist/cjs/v4.js` in my `node_modules` looks like:


    if (native_js_1.default.randomUUID && !buf && !options) {
        return native_js_1.default.randomUUID();
    }
    // ...unreachable
}


It looks like that if condition is causing all the slowdown. If I replace it with `true`, it runs as fast as native. I can confirm the code below is unreachable, as throwing an exception does nothing down there.

If I leave _any_ of what's in that if condition in there, it still is slow. Only if I replace it with true does it run fast.

I can reproduce _also_ if `type: module` is set in `package.json` (after modifying the imports), however, obviously I need to edit the `esm/v4.js` in `node_modules`.

What could be happening here?
```

### Environment

```
  System:
    OS: macOS 15.6
    CPU: (10) arm64 Apple M4
    Memory: 86.84 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.16.0 - ~/.nvm/versions/node/v22.16.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v22.16.0/bin/npm
    pnpm: 10.13.1 - ~/.nvm/versions/node/v22.16.0/bin/pnpm
  Browsers:
    Chrome: 138.0.7204.184
    Safari: 18.6
  npmPackages:
    uuid: ^11.1.0 => 11.1.0
```

broofa:
Moving tail-code in v4() to a separate function appears to enable some interesting compiler optimizations.  `v4()` gets a 3X boost in Node.  In Chrome, there's a small (7% or so) boost, but not enough to be noteworthy.

@pnappa, can you pull this branch and comment here with your results for `npm run test:benchmark`?

Fixes #892

> [!NOTE]
> node@18, 20, 22, and 24 all show similar results with this change

### Benchmarks
node@24 before:
```
uuid.v4() using crypto.randomUUID x 7,342,641 ops/sec ±0.27% (101 runs sampled)
uuid.v4() w/out crypto.randomUUID x 3,301,438 ops/sec ±0.25% (98 runs sampled)
crypto.randomUUID() x 17,718,991 ops/sec ±0.62% (95 runs sampled)
```

node@24 after:
```
uuid.v4() using crypto.randomUUID x 22,060,386 ops/sec ±0.60% (94 runs sampled)
uuid.v4() w/out crypto.randomUUID x 7,832,501 ops/sec ±0.35% (96 runs sampled)
crypto.randomUUID() x 17,428,071 ops/sec ±0.57% (97 runs sampled)
```

Chrome@138 before:
```
uuid.v4() using crypto.randomUUID x 647,225 ops/sec ±7.95% (57 runs sampled)
benchmark.js:84 uuid.v4() w/out crypto.randomUUID x 952,060 ops/sec ±0.59% (67 runs sampled)
benchmark.js:84 crypto.randomUUID() x 620,774 ops/sec ±5.69% (58 runs sampled)
```

Chrome@138 after:
```
benchmark.js:84 uuid.v4() using crypto.randomUUID x 699,045 ops/sec ±3.02% (49 runs sampled)
benchmark.js:84 uuid.v4() w/out crypto.randomUUID x 966,959 ops/sec ±1.68% (68 runs sampled)
benchmark.js:84 crypto.randomUUID() x 638,242 ops/sec ±3.49% (58 runs sampled)
```

### DIFF
--- a/src/v4.ts
+++ b/src/v4.ts
@@ -51,9 +51,8 @@ function v4<TBuf extends Uint8Array = Uint8Array>(
     return native.randomUUID();
   }
 
-  // Putting tail-code here in a separate function (as opposed to inline in this
-  // function) allows for compiler optimizations in the common case where
-  // we never get this far (i.e. when native.randomUUID() is called, above).
+  // Putting tail-code that could just go inline here in a separate function
+  // allows for compiler optimizations that dramatically improve performance.
   //
   // REF: https://github.com/uuidjs/uuid/issues/892
   return _v4(options, buf, offset);


pnappa:
Just tested it, with the same configuration as #892 - and I do also observe the same improvement. `v4()` is now comparable to native. Great work!

How'd you come up with the refactor? A hunch about making the else-case look less desirable to the optimiser?

broofa:
> How'd you come up with the refactor? A hunch about making the else-case look less desirable to the optimiser?

Mostly trial-and-error.  The presence of the tail-code was obviously having an impact, even though it was never getting used.  If we wanted to keep the behavior of `v4()` the same but somehow change the optimiser behavior there weren't a lot of options for other ways to structure the code.  I figured it was worth a shot... and it worked.  🤷

masoud-msk:
Nice optimization! But a question:
How is it `uuid.v4()` is faster than `crypto.randomUUID()` when the former is calling the latter?
Doesn't it mean that the tests are not valid?
Could you come up with any other explanation to this?

PR #895, opened 2025-08-12
feat: add type for uuid

legend80s:


LinusU:
What is the use case for this?

I have a feeling that this will increase TypeScript compile times? 🤔

PR #916, opened 2025-11-12
Fix Array Out-of-Bounds and Optimize MD5 Algorithm Implementation

gaoyia:
Fix Array Out-of-Bounds and Optimize MD5 Algorithm Implementation

log-position:
```ts
    c = md5ff(c, d, a, b, x[i + 14], 17, -1502002290);
    console.log(x, i + 15)
    b = md5ff(b, c, d, a, x[i + 15], 22, 1236535329);  // ←  Array Out-of-Bounds
```

test-code:
```ts
const str = '123456'

const arr= new Uint8Array(str.length);
for (let i = 0; i < str .length; i++) {  
  arr[i] = str.charCodeAt(i);
}
```
output:
```js
Uint32Array(15) [
  875770417, 8402485,  0,
          0,       0,  0,
          0,       0,  0,
          0,       0,  0,
          0,       0, 48
] 15
```

### DIFF
--- a/src/md5-browser.ts
+++ b/src/md5-browser.ts
@@ -32,8 +32,8 @@ function uint32ToUint8(input: Uint32Array) {
  */
 function wordsToMd5(x: Uint32Array, len: number): Uint32Array {
   /* append padding */
-  x[len >> 5] |= 0x80 << (len % 32);
-  x[((len + 64 >>> 9) << 4) + 14] = len;
+  x[len >> 5] |= 0x80 << len % 32;
+  x[(((len + 64) >>> 9) << 4) + 14] = len;
 
   let a = 1732584193;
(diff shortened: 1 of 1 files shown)


broofa:
Thanks for taking the time to put this together.  I'm currently disinclined to take this for the following reasons:

Regarding the out-of-bounds issue:  Please create an issue with an MRE (e.g. a codesandbox or stackblitz project) that shows how calling into the **public** `uuid` API - i.e. the `v3()` method - will trigger this out-of-bounds issue.  (If the issue can't be triggered that way, I don't consider it a bug.)

Regarding performance:  Do you have a use case where v3() performance is a bottleneck?  If so, I'd love to hear it.  Otherwise I don't consider performance a priority for this project.

Issue #856, opened 2025-01-07
[BUG] 11.0.4 "[ERROR] TS2315: Type 'Uint8Array' is not generic. [plugin angular-compiler]" build error in Angular 18 and 19


Servonius:
### Before you begin...

- [X] I have searched the existing issues
- [X] I am not using version 13.x of node  (if so, please upgrade)

### Description of the problem

The latest patch version 11.0.4 is no longer usable with Angular 18 and 19 because the patch version is requiring Typescript 5.7 which is not supported by Angular 18 and 19.

It worked with previous 11.0.X versions.

It causes the following build error:

```
✘ [ERROR] TS2315: Type 'Uint8Array' is not generic. [plugin angular-compiler]
    node_modules/uuid/dist/cjs/parse.d.ts:1:38:
      1 │ declare function parse(uuid: string): Uint8Array<ArrayBuffer>;
        ╵                                       ~~~~~~~~~~~~~~~~~~~~~~~
✘ [ERROR] TS2315: Type 'Uint8Array' is not generic. [plugin angular-compiler]
    node_modules/uuid/dist/cjs/v35.d.ts:2:52:
      2 │ ...clare function stringToBytes(str: string): Uint8Array<ArrayBuffer>;
        ╵                                               ~~~~~~~~~~~~~~~~~~~~~~~
✘ [ERROR] TS2315: Type 'Uint8Array' is not generic. [plugin angular-compiler]
    node_modules/uuid/dist/cjs/v35.d.ts:6:165:
      6 │ ...Uint8Array, offset?: number): string | Uint8Array<ArrayBufferLike>;
        ╵                                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

andyf0x:
Same here, but for Angular 16. Fixing on 11.0.3 has resolved the issue.

broofa:
For future reference, this is exactly why we have this in the issue template:

>  "IMPORTANT: Failure to provide a [Minimal, Complete, and Verifiable](https://stackoverflow.com/help/minimal-reproducible-example) example will result in this issue being closed without further review."

[...]

PR #927, opened 2026-03-16
chore: upgrade TypeScript and fixup types

jamiebuilds:
I was trying to make use of `uuid.parse()` but the return type was `Uint8Array` instead of the `Uint8Array<ArrayBuffer>` type that `Uint8Array.of()` actually returns.

In order to keep it compatible with different versions of TypeScript'd definitions I've used:

```ts
type NonSharedArrayBuffer = ReturnType<typeof Uint8Array.of>;
```

Which then should be the correct type no matter what version of TypeScript someone is on

I also upgraded the version of TypeScript to double check that nothing would break, and I just had to fix one small type in `v35()`


src/parse.ts
```diff
@@ -1,6 +1,9 @@
 import validate from './validate.js';
 
-function parse(uuid: string) {
+// Workaround for different versions of TypeScript definitions
+type NonSharedArrayBuffer = ReturnType<typeof Uint8Array.of>;
```
broofa: Move to types.ts so it's not copy/pasted across files.

broofa:
Related to #856

broofa (changes requested):
~~[Trying out using copilot to fix some of these issues.  Not sure how that's going to work out, but we'll figure it out.]~~ Nevermind.  Copilot doesn't work on x-repository requests.

Thanks for the contribution.  The `NonSharedArrayBuffer` type seems like a nice solution to the TS version problem, but something here broke the `npm run test:node` tests.  Those will need to get fixed.

broofa (changes requested):

PR #935, opened 2026-04-06
fix!: expect `crypto` to be global everywhere (requires node@20+)

### LINKED ISSUE #891: [after node@18 is dropped] Combine rng.ts + rng-browser.ts
Node provides `crypto` as a global in node@20+, which should allow us to leverage `crypto.getRandomValues()` for both node and browser.

See also #890

broofa:
Now that `crypto` is globally defined everywhere (all browser environments and node@20+), we can clean up the feature detection logic we had around it.

* Remove `typeof crypto` feature-detection
* Remove `rng-browser.ts`
* Remove `native*.ts`

Closes #891


realityking:
Very cool. Thank you!

jimmycallin:
hi! will this update break usage of uuid in non-secure browser contexts? we are relying on uuid over crypto due to crypto.randomUUID not being available in non-https environments.

realityking:
I haven’t tested it myself but if I’m reading the code right it’ll fallback to crypto.randomValues() if crypto.randomUUID() is not present.

jimmycallin:
> I haven’t tested it myself but if I’m reading the code right it’ll fallback to crypto.randomValues() if crypto.randomUUID() is not present.

i just tested this myself, and in contrast to my assumptions `crypto.randomUUID` is simply not available in insecure contexts. i assumed it would throw an error. so this should be fine. thank you!

broofa:
> if I’m reading the code right it’ll fallback to crypto.[getR]andomValues()

This is correct.  FWIW, we test all of the `examples/browser-*` samples in non-secure contexts as part of our CI checks.  E.g...

CleanShot 2026-05-04 at 05 57 00@2x

PR #937, opened 2026-04-19
fix: throw range error for invalid buf and offset args

broofa:


LinusU (approved):
👍

broofa:
Closed this and, instead, went through the "fork private, temporary repository" flow recommended in the https://github.com/uuidjs/uuid/security/advisories/GHSA-w5hq-g745-h8pq advisory.

PR #961, opened 2026-06-15
fix: add types condition to node export for moduleResolution bundler

XananasX7:
## Problem

`uuid@14.0.0` defines `exports["."].node` as a plain string pointing to `./dist-node/index.js`, but omits a `types` field for that condition. `publint` warns that TypeScript consumers using `moduleResolution: "bundler"` cannot resolve types when importing via the `node` export condition.

**Reproduction:**
```shell
npm pack uuid@14.0.0 --silent >/dev/null
npx publint@0.3.21 run uuid-14.0.0.tgz --level warning
```

Actual output:
```
pkg.exports["."].node types is not exported. Consider adding pkg.exports["."].node.types: "./dist/index.d.ts"
to be compatible with TypeScript's "moduleResolution": "bundler" compiler option.
```

## Fix

Expand the `node` condition from a plain string to an object with an explicit `types` field:

```diff
-"node": "./dist-node/index.js",
+"node": {
+  "types": "./dist/index.d.ts",
+  "default": "./dist-node/index.js"
+},
```

## Verification

After patching, packing, and running publint on the tarball:
```
Running publint v0.3.21 for uuid-14.0.0.tgz...
All good!
```

No runtime behaviour is changed — the fix only adds metadata for TypeScript tooling.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
       - run: npm ci
       - run: npm run lint
       - run: npm run docs:diff
+      - run: npx -y publint --strict
 
   test:
     runs-on: ubuntu-latest
--- a/package.json
+++ b/package.json
@@ -26,7 +26,10 @@
   "types": "./dist/index.d.ts",
   "exports": {
     ".": {
-      "node": "./dist-node/index.js",
+      "node": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist-node/index.js"
+      },
       "default": "./dist/index.js"
     },
     "./package.json": "./package.json"


broofa:
Please add a CI check for this to `.github/workflows/ci.yml`.

Specifically, I think we'll want something like this (at line 24)...
```yaml
      - run: npm ci
      - run: npm run lint
      - run: npm run docs:diff
      - run: npx -y publint --strict  # <--- ADD THIS
```

XananasX7:
Thanks for the suggestion @broofa! Makes sense — adding `publint --strict` as a CI gate will catch any future export misconfiguration before it ships.

I'll add the `publint` step to `.github/workflows/ci.yml` at line 24. Could you confirm that `npx -y publint --strict` is sufficient, or should it be pinned to a specific version?

XananasX7:
Done — added `npx -y publint --strict` to the `lint` job in `.github/workflows/ci.yml` right after `npm run docs:diff`. Didn't pin to a specific version since `npx -y` will always pull the latest, which seems consistent with how the project uses `npx` elsewhere. Happy to pin it to a specific semver if you'd prefer.

broofa (approved):

PR #962, opened 2026-06-19
docs: add Deno to install instructions

bartlomieju:
👋 Bartek from the Deno team here.

The README shows `npm install uuid`. Since Deno is a drop-in replacement for npm these days, this adds a Deno line alongside it:

```sh
deno install uuid
```

Docs-only change (uuid resolves and installs the same way under Deno). Totally fine to close this if you'd rather keep it npm-only. Happy to adjust.

_Disclosure: I work on Deno, so I have an interest here. The goal is parity with the package managers you already list, not promotion. This PR was prepared with AI assistance under human supervision: I review every change myself and personally respond to any comments or review feedback._

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ For the creation of [RFC9562](https://www.rfc-editor.org/rfc/rfc9562.html) (form
 
 ```shell
 npm install uuid
+# or
+deno install uuid
 ```
 
 **2. Create a UUID**


broofa:
> The goal is parity with the package managers you already list, not promotion.

So why no instructions for `yarn`, `pnpm`, and `bun` as well? 😉

Anyone using a non-`npm` package manager will know what the default incantation is for installing a package.  There's no need to clutter up the README with that information.  And mentioning `deno` (or `bun` or `yarn` or `pnpm`) implies we test and support those environments, which we don't.

(And why care about promoting `deno` here when there's [`@std/uuid`](https://docs.deno.com/runtime/reference/std/uuid/)?)

PR #965, opened 2026-06-22
fix(v7): align default seq formula in v7Bytes with updateV7State

JSap0914:
## Problem

When `v7()` is called **with options but without an explicit `seq`**
(e.g. `v7({ random: rnds, msecs: t })`), `v7Bytes` computes the
default sequence number as:

```ts
seq ??= ((rnds[6] * 0x7f) << 24) | (rnds[7] << 16) | (rnds[8] << 8) | rnds[9];
```

The `* 0x7f` (multiply-by-127) is **inconsistent** with the formula
used by `updateV7State` when time advances:

```ts
state.seq = (rnds[6] << 23) | (rnds[7] << 16) | (rnds[8] << 8) | rnds[9];
```

For even values of `rnds[6]`, the product overflows the low byte
(e.g. `rnds[6] = 0x02` → `2 × 127 = 254 = 0xFE`), and
`0xFE << 24` is the **negative** 32-bit integer `0xFE000000`.
The subsequent unsigned right-shift `seq >>> 28` then extracts
`0xF` instead of the near-zero value that `updateV7State` would
produce for the same bytes, so the two code paths generate different
UUIDs for identical inputs:

```
// rnds[6] = 0x02
updateV7State path: 017f22e2-79b0-701c-8c63-100c0c07398f
v7Bytes default:    017f22e2-79b0-7fec-8c63-100c0c07398f
                                  ^
```

## Fix

Replace `(rnds[6] * 0x7f) << 24` with `rnds[6] << 23` to match
`updateV7State`, keeping `seq` non-negative (range 0 – 0x7F800000)
and the two code paths consistent.

## Testing

A new test `'default seq (no explicit seq option) is consistent with updateV7State formula'`
was added to `src/test/v7.test.ts`. It **fails** before the fix
(`017f22e2-79b0-7fec-…` ≠ `017f22e2-79b0-701c-…`) and **passes**
after.

All 77 existing tests continue to pass.

> AI-assisted contribution.

### DIFF
--- a/src/test/v7.test.ts
+++ b/src/test/v7.test.ts
@@ -173,7 +173,7 @@ describe('v7', () => {
         now: 2,
         expected: {
           msecs: 2, // time interval should update
-          seq: 0x6c318c4, // sequence should be randomized
+          seq: 0xcc318c4, // sequence should be randomized
         },
       },
       {
@@ -301,11 +301,10 @@ describe('v7', () => {
   test('default seq (no explicit seq option) is consistent with updateV7State formula', () => {
     // When v7() is called with random bytes and msecs but no explicit seq, the
     // default seq should use the same formula as updateV7State() –
-    // (rnds[6] << 23) | (rnds[7] << 16) | (rnds[8] << 8) | rnds[9].
+    // ((rnds[6] & 0x7f) << 24) | (rnds[7] << 16) | (rnds[8] << 8) | rnds[9].
     //
-    // Regression: the formula was `(rnds[6] * 0x7f) << 24`, which overflows
-    // for even values of rnds[6] (e.g. rnds[6]=0x02 gives 0xFE000000, a
-    // negative 32-bit integer), producing a seq inconsistent with
+    // Regression: the formula was `(rnds[6] * 0x7f) << 24`, which should have
+    // been `(rnds[6] & 0x7f) << 24`, producing a seq inconsistent with
     // updateV7State.
     const rnds = Uint8Array.of(
       0x02,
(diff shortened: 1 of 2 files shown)


broofa (changes requested):
Hi, and thanks for catching this.  

You're correct that the default value of `seq` should be initialized the same way in both places.  But the proposed fix is not correct.  Please make the following changes:

  1.  Centralize the logic for converting `rnds` to `seq` by adding this method to v7.ts:
```ts
function v7Sequence(rnds: Uint8Array) {
  return ((rnds[6] & 0x7f) << 24) | (rnds[7] << 16) | (rnds[8] << 8) | rnds[9];
}
```
  2. In `updateV7State` set `state.seq = v7Sequence(rnds)`
  3. In `v7Bytes`, initialize `seq ??= v7Sequence(rnds)`
  4. Update your new unit test accordingly

---
Explanation:

Image

`rnds[6] * 0x7f` is a typo that should be `rnds[6] & 0x7f`.   That this still injected `rnd[6]` bits into `seq` is just a lucky result of two's-complement notation. 😆 

FWIW, the reason for not setting the high-order bit of `seq` is to guard against rollovers in subsequent `v7()` calls, [as mentioned in the RFC](https://www.rfc-editor.org/rfc/rfc9562.html#section-6.2-6):

> ... The remaining most significant, leftmost counter bit is initialized as zero for the sole purpose of guarding against counter rollovers.

Lastly, while `rnds[6] << 23` avoids setting the high-order bit of `seq`, it ends up setting `seq: bit 23` to the `OR`ed result of `rnds[6]:bit 0` and `rnds[7]: bit 7`.  There's nothing overtly wrong with that but it's not ideal.

broofa (approved):

danissulih1-dev (commented):
// rnds[6] = 0x02
updateV7State path: 017f22e2-79b0-701c-8c63-100c0c07398f
v7Bytes default:    017f22e2-79b0-7fec-8c63-100c0c07398f
                                  ^

PR #971, opened 2026-07-23
docs: clarify namespace arrays vs stringify validation

### LINKED ISSUE #970: Q: Clarify meaning of "UUID" (namespace vs stringify)
Hi, thanks for the UUIDs!

I have a question. I found out the hard way that `stringify` doesn't stringify any old length-16 `UInt8Array`. But I believe `v3` and `v5` do accept any such array, for the namespace. The docs for `v5` say

> namespace   |    String | Array[16] Namespace UUID

but looking through your past gh issues I understand you consider "UUID" to mean "valid UUID", so I though I'd ask for clarification. Are `v3` and `v5` meant to accept any `UInt8Array`, whether a UUID or not?

Hashim1999164:
## Summary
- Clarify that `v3()` / `v5()` namespace `Array[16]` values may be any 16 bytes, while string namespaces must be valid UUIDs.
- Note that `stringify()` requires RFC-valid UUID bytes (version and variant), unlike namespace arrays.
- Document that `validate()` is string-only and that `stringify()` can be used to check byte arrays.

Fixes #970

## Test plan
- [ ] Confirm README_js.md and generated README.md notes match and read clearly
- [ ] Spot-check API docs for `stringify`, `v5`, and `validate` sections


broofa:
https://github.com/uuidjs/uuid/issues/970#issuecomment-5191620776

PR #973, opened 2026-08-08
fix(v1): set the multicast bit on v1Bytes's own randomly-generated node

rajanpanth:
## Summary
RFC 9562 §6.10 requires that a randomly-generated node ID have its multicast bit set. `updateV1State()` already does this correctly for the node it randomizes and caches, but that's only reached by plain `v1()` calls with no options. Every other path that generates its own random node computes it inline in `v1Bytes()` via `node ??= rnds.slice(10, 16)`, without the same bit set:

- `v1(options)` whenever `options` omits `node`
- `v6()` **unconditionally** — `v6()` always defers its random node to `v1Bytes()` rather than `updateV1State()`, with or without options

Since `v6()` is the RFC-recommended, time-ordered generator, this isn't a rare edge case:

```
No-options v1():   multicast bit set 2000 / 2000   (100%, correct)
v1({ msecs }):     multicast bit set  972 / 2000   (~49%, wrong)
v6() no options:   multicast bit set 1030 / 2000   (~52%, wrong)
v6({ msecs }):     multicast bit set 1000 / 2000   (50%,  wrong)
```

The fix already exists 8 lines away in the same file (`updateV1State`) with a comment citing the same RFC anchor — it just never got propagated to `v1Bytes`'s own default.

## Changes
- `src/v1.ts`: set the multicast bit when `v1Bytes` generates its own random node, mirroring `updateV1State`'s existing pattern. Only applies when no `node` option was explicitly supplied — an explicit node (e.g. a real MAC address) is left untouched, since forcing the bit there isn't warranted by the RFC clause and could be surprising for a caller-supplied identifier.
- `src/test/v1.test.ts`, `src/test/v6.test.ts`: one test each, confirmed to fail on the prior code (~50% pass rate over repeated runs) and pass reliably with the fix.

## Testing
- Full suite: 82/82 passing (up from 80).
- `biome check`, `tsc --noEmit` clean.
- Confirmed empirically before/after with a standalone repeated-trial script (500+ trials per path).


broofa (approved):
Great catch!  Thanks for the contribution.

PR #974, opened 2026-08-17
docs: point the Node support permalink at the CI version matrix

FirmaSpring:
## Problem

The Node support paragraph in the README permalinks into `.github/workflows/ci.yml#L26-L27` as the Node version matrix. On current `main` those lines are the start of the `test` job (`runs-on` / `strategy`), not the versions.

The matrix is now:

```
node-version: [20.x, 22.x, 24.x]
```

at line 31.

## Solution

Point the permalink at `#L31`, the line that actually lists the tested Node versions.

`README.md` is regenerated from `README_js.md`; both files are updated together so `docs:diff` stays clean.

## Verification

```
curl -sL https://raw.githubusercontent.com/uuidjs/uuid/main/.github/workflows/ci.yml | nl -ba | sed -n '26,31p'
# 26   test:
# 27     runs-on: ubuntu-latest
# 31         node-version: [20.x, 22.x, 24.x]
```


broofa (approved):
Looks good.  Thanks for the contribution.

PR #976, opened 2026-09-15
fix: type version() to accept unknown, as the runtime does

hxperl:
`version()` throws `TypeError` for any non-UUID, as the README documents, so `string` is narrower than its real contract.

The repo's own tests show it. `validate.test.ts` and `version.test.ts` iterate the same `TESTS` array — which holds `undefined`, `null`, `123`, `/regex/`, `false` — but only `version.test.ts` needs a suppression:

```ts
// @ts-expect-error testing invalid input
const actualVersion = version(value);
```

And the README's per-version check fails to compile. Added to `examples/typescript/index.ts`, on main:

```
index.ts(8,47): error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string'.
```

`validate()` already took `unknown` after #606; this does the same for `version()`, so the suppression becomes unused.

I considered a `uuid is string` predicate on `validate()` instead, but it narrows the false branch too — `if (!validate(s)) throw Error(s.toUpperCase())` breaks with `s` as `never`. Happy to switch if you'd rather.

`npm test` 82/82, `examples:node:typescript:test` and biome clean, macOS arm64.

Disclosure: prepared with Claude Code; diff, reproduction and test runs verified by me.


broofa:
I don't see changing this check from static to dynamic as a good thing.  Stricter types are generally considered better.

> And the README's per-version check fails to compile. Added to examples/typescript/index.ts, on main:
> 
> index.ts(8,47): error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string'.

The README code is JS, not TS, so this isn't _technically_ a problem in the README, as written.  That said, I see your point.  Converting that example to TS is... "problematic". 😄 

I think there's actually a simpler/better solution here, which is to add a type-guard for the `uuid` arg to the `validate()` method, as follows:

```ts
function validate(uuid: unknown): uuid is string {
  return typeof uuid === 'string' && REGEX.test(uuid);
}
```

Can you give that a try and let me know what you think?

broofa (changes requested):
I don't see changing the `string` type-check from static to dynamic as a good thing.  Stricter types are generally considered better.  And the existing type semantics for `validate` ("Validate this _thing_ is a UUID") and `version` ("Give me the version of this  _UUID_") seem appropriate.

> And the README's per-version check fails to compile. Added to examples/typescript/index.ts, on main:
> 
> index.ts(8,47): error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string'.

The README code is JS, not TS, so this isn't _technically_ a problem in the README, as written.  That said, I see your point.  Converting that example to TS is... "problematic". 😄 

I think there's actually a simpler/better solution here, which is to declare `validate` as a type-guard for the `uuid` arg, as follows:

```ts
function validate(uuid: unknown): uuid is string {
  return typeof uuid === 'string' && REGEX.test(uuid);
}
```

Can you give that a try and let me know what you think?

PR #977, opened 2026-09-15
feat!: add type assertion to `validate`

icopp:
The type assertion automatically narrows down the implicit typing of whatever value it is given to the templated string type. Most users won't care about the specific templated detail and will just treat it the same as `is string`, but the templated type is the closest TypeScript can get to encoding the literal pattern of a UUID in types.

This is a breaking change because depending on strictness, this may make existing code error if it's doing something invalid after the `validate` check.


src/validate.ts
```diff
@@ -1,6 +1,6 @@
 import REGEX from './regex.js';
 
-function validate(uuid: unknown) {
+function validate(uuid: unknown): uuid is `${string}-${string}-${string}-${string}-${string}` {
```
broofa, line 3:
```suggestion
function validate(uuid: unknown): uuid is string {
```

`uuid` uses `string` for UUID (string) types.  Ad-hoc template types like this are more trouble than they're worth (at least in the context of `uuid`).

broofa (changes requested):
Re: how this affects the semver, [ChatGPT suggests this should be a patch](https://chatgpt.com/s/t_6aa9d4708d00819185e78cf9666bdc22).

src/validate.ts
icopp: I've updated it that way

icopp:
> Re: how this affects the semver, [ChatGPT suggests this should be a patch](https://chatgpt.com/s/t_6aa9d4708d00819185e78cf9666bdc22).

If it was something internal with controlled consumers, that would make sense to me, but I'd bet anything there are codebases out there right now that this would cause failed builds for.

broofa:
> If it was something internal with controlled consumers, that would make sense to me, but I'd bet anything there are codebases out there right now that this would cause failed builds for.

Can you give me an example of how this would break existing code?  The type is narrowed (`unknown` or `any` -> `string`).  I'm struggling to come up with an example of consumer code where this would be an issue.

